PR #58, opened 2026-07-16
feat: Add District-Level Set (Batch) Management featureure

kudimirohith-bit:
## What
Adds District-Level Set (Batch) Management: District/Block Organizers can group
students into Sets, bulk-generate personalized question papers for every student
in a Set, merge them into one printable package (cover page + papers in student
order), and track each Set through its status lifecycle (Created → ... →
Evaluation Completed).

## Why
Printing/generating papers one student at a time doesn't scale for district-wide
assessments (thousands of students). This lets an organizer manage generation,
printing, and tracking at the Set level while each student still gets their own
personalized paper — personalization logic is untouched.

## New endpoints
- POST   /api/sets
- GET    /api/sets            (filter by setId, assessmentName, schoolId, classGroup, status)
- GET    /api/sets/:id
- PATCH  /api/sets/:id/status (enforces forward-only lifecycle transitions)
- POST   /api/sets/:id/generate
- GET    /api/sets/:id/generate/:jobId/progress
- GET    /api/sets/:id/download

## Access control
All /api/sets* routes are restricted to DISTRICT_ADMIN, BLOCK_ADMIN, SUPERADMIN,
and force-scoped by the caller's districtCode/blockCode — a District Admin can
only see/act on Sets belonging to schools in their own district, regardless of
query params.

## Known limitations (by design, not oversight)
- Set generation jobs are tracked in-memory (same pattern as the existing
  /api/diagnostic/bulk jobs) and are lost on server restart.
- The download endpoint accepts the auth token as a query param as well as a
  header, since the frontend triggers download via a plain link click (browsers
  don't attach custom headers to navigation). This matches the app's existing
  token scheme (a plain identifier, not a signed/expiring credential), so this
  isn't a new class of exposure — but it does mean the token can appear in
  browser history/server logs. Flagging for awareness.

## Testing done
- Manual end-to-end: create → search/filter → generate → poll progress →
  download → status transitions (valid + rejected invalid transitions).
- Verified role/scope enforcement: non-organizer roles get 403; a district admin
  cannot see or act on another district's Sets.
- `tsc --noEmit` clean on both backend and frontend.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -22,9 +22,9 @@
     "express-validator": "^7.2.1",
     "http-status": "^2.1.0",
     "jsonwebtoken": "^9.0.2",
-    "mongoose": "^8.9.5",
     "jszip": "^3.10.1",
     "mongodb": "^6.12.0",
+    "mongoose": "^8.9.5",
     "pdf-lib": "^1.17.1",
     "puppeteer": "^25.3.0"
   },
@@ -37,6 +37,7 @@
     "concurrently": "^10.0.3",
     "esbuild": "^0.25.0",
     "tsx": "^4.21.0",
-    "typescript": "~5.8.2"
+    "typescript": "~5.8.2",
+    "vite": "^8.1.5"
   }
 }
--- a/backend/package.json
+++ b/backend/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "dev:old": "tsx src/index.ts",
-    "dev:new": "tsx src/server.ts",
+    "dev:new": "PORT=3001 tsx src/server.ts",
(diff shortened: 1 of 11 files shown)


jgupta05072003-code:
Thank you for this contribution.

We have not yet done a full review of this pull request, since it currently has a merge conflict with the main branch. Kindly pull the latest main branch into your branch and resolve the conflict at your end.

Also, we could not find a linked issue in the pull request description. Kindly link the issue this pull request is meant to resolve, using "Closes #N", so that it can be tracked properly. If there is no existing issue for this work, kindly open one describing the motivation before we proceed with a review.

Once the conflict is resolved and the issue is linked, we will proceed with a full review.

PR #108, opened 2026-07-31
feat: Add Printable Weekly Student and Top-5 Achievement Certificates

intelliraff:
## **Summary**

This PR introduces printable weekly student certificates and Top-5 achievement certificates for Teacher and Volunteer users.

## **Changes Made**

- Added weekly certificates to individual student profiles
- Added Top-5 achievement certificates to the performance dashboard
- Added class-specific normal certificate designs
- Included student details, current FLN level, learning streak, class rank, issue date, and teacher name
- Added the FLN Portal heading to Top-5 certificates
- Implemented accurate A4 landscape printing at `297mm × 210mm`
- Restricted certificate access to Teacher and Volunteer users
- Preserved existing student-data loading and frontend functionality

## **Benefits**

- Recognizes weekly student progress and achievement
- Allows teachers and volunteers to print professional certificates
- Highlights Top-5 student performance
- Produces consistent, print-ready A4 certificates
- Prevents unauthorized certificate access from School-level accounts
- Uses existing student roster data without requiring additional API calls

## **Testing**

- Verified normal and Top-5 certificate rendering
- Verified A4 landscape print dimensions
- Verified normal certificates print without top offset
- Verified School users cannot access certificate features
- Verified Teacher and Volunteer access remains available
- Verified frontend TypeScript checks pass
- Verified the frontend production build succeeds


jgupta05072003-code:
Hi @intelliraff — this has been open since 31 July with no reply from us at all. Forty-three days of silence on real work is not acceptable and I am sorry. Here is a proper review.

Printable weekly reports and a top-5 achievement view are worth having — teachers asking for something they can put on a wall or send home is a real request, and 1,459 lines is a serious piece of work.

There are two things to resolve before it can merge. I want to be clear that the second is a policy that was tightened *after* you opened this, so it is not something you could have known.

## 1. The demo-data fallbacks cannot ship

The PR adds five mock constants and falls back to them whenever the API returns nothing:

```js
const REPORTS_MOCK: EvaluationReport[] = [ ... ]
const TEACHERS_MOCK = [ ... ]
const WORKSHEETS_MOCK = [ ... ]
const ATTENDANCE_MOCK = [ ... ]
const SYSTEM_LOGS_MOCK = [ ... ]

const reportsList = apiReports.length > 0 ? apiReports : REPORTS_MOCK;
const teachersList = apiTeachers.length > 0 ? apiTeachers : TEACHERS_MOCK;
```

Two consequences worth spelling out:

```js
<MetricCard title="Total Worksheets" value={WORKSHEETS_MOCK.length} ... />
const att = ATTENDANCE_MOCK.find(a => a.student === s.name);
```

The first shows an **invented number in a metric card** with no indication it is fake. The second **attaches fabricated attendance records to real children's names**. On a printable report going home to a parent, that is the most damaging place a placeholder can end up.

**This is the tightened policy, recorded in #449**: the platform must never present invented data as real. An empty state saying "No worksheets yet" is always better than a plausible fake number, because a teacher cannot tell the difference. It is being actively enforced — PR #471 is currently removing this same fallback pattern from other panels, so merging this would reintroduce what we are in the middle of taking out.

**What we need:** drop the five mock constants and render genuine empty states when the API returns nothing. Keeping the mocks in a test fixture is fine; they just must not reach a user.

## 2. 1,351 lines into `PanelViews.tsx`

That file is **160 lines on `main`**, so this makes it roughly nine times larger. It is a routing switch, and we are trying to keep it that way — the same monolith problem in Tenali's `App.jsx` is now a multi-issue programme to unpick.

**What we need:** move each new panel into its own component under `frontend/src/components/panels/`, matching the ~20 that already live there, and leave `PanelViews.tsx` with just the `if (panel === 'x') return <XPanel/>` lines. That is a mechanical change and it will make the PR far easier to review — most of the diff becomes new files rather than one enormous one.

## Smaller notes

- You add six panel keys (`certificates`, `diagnostic_test`, `adaptive_test`, `test_history`, `worksheets`, `student_profile`). Please check each is reachable from the sidebar for the intended roles — we recently found two nav items pointing at panels that did not exist, rendering blank screens (#446).
- `frontend/src/index.css` (+108) — if these are print styles, worth saying so; print CSS is easy to break accidentally later.
- You will need a rebase; `main` has moved a fair way since July.

## If you have moved on

Completely understandable after this long, and no hard feelings — just say so. But the underlying feature is wanted, and with the mocks removed and the panels extracted I think this lands. **If you pick it back up, I will review within a day** — you should not have to wait again.

Sorry once more for the delay. That is a failure of our process, not of your contribution.

intelliraff:
Thank you for the review feedback. I’ve made all the requested changes, including removing the mock data, extracting the panels into separate components, checking the navigation, and updating the print styles.

video demo: https://github.com/user-attachments/assets/e5530e5b-78d6-4feb-a3fc-3997e0ce6172

@jgupta05072003-code

PR #59, opened 2026-07-16
feat: Add Intervention Dashboard

jsumanigoud-svg:
## Summary
- Added Intervention Dashboard under Students
- Integrated frontend with backend intervention dashboard API
- Displayed intervention recommendations in a table
- Added navigation to access the dashboard

## Testing
- Verified API endpoint `/api/interventions/dashboard`
- Verified dashboard loads intervention data successfully
- Tested using teacher login

## Related Issue
Closes #514

PR #158, opened 2026-08-08
Backend hardening: rate-limit ICR, fix XSS/data-export gaps, consolidate LLM providers, drop 49MB dead weight

tanvishdesai:
## Summary

A batch of independently-verified fixes from an audit of the backend, delivered as one PR with a real test suite (there wasn't one before). Every finding below was confirmed by reading the actual code paths, not inferred from symptoms — file:line references are in the commit messages.

- **No rate limit / no size cap on the cloud OCR endpoints.** `POST /api/icr/evaluate-cloud` and `/api/icr/evaluate-pdf` were reachable by any authenticated account (any role) with no throttling and no payload cap. `evaluate-cloud` forwards to a paid external API (Google Vision / MiniMax / OCR.space) using server-held keys, and `evaluate-pdf` spawns a 60s local subprocess per call — both were direct, unbounded cost/CPU exposure. Added the same `express-rate-limit` pattern already used for login (20 req/15min), and the same 8MB decoded-size cap `/api/icr/filter` already enforced, now applied consistently across all three ICR upload endpoints. Also fixed a temp-file leak: `evaluate-pdf` only cleaned up its scratch file on the success path, so any exception after the file was written left it on disk permanently.
- **HTML injection into a server-rendered PDF.** `paperGenerator.ts` interpolated `studentName`/`studentId` unescaped into HTML fed to `printPage.setContent()` — a real, network-capable headless Chrome instance. A crafted name could execute inside that Chrome instance server-side, not just corrupt the PDF visually. Added `escapeHtml()` and applied it at both interpolation points; verified live against the real Puppeteer pipeline with an `` payload as the student name (no injection, no crash, PDF renders the literal text).
- **Unbounded `?all=1` export on `/api/students`.** `all=1` set `limit=0`, which the query builder only forwarded `if (limit > 0)` — meaning `all=1` actually meant "no limit passed to the query at all," pulling the full 86k+ row table in one response for any account with broad-enough role scope. Introduced a 10,000-row hard ceiling that applies even when `all=1` is requested, preserving `all=1`'s actual purpose (bypass the default 1000-row page) without the unbounded-dump behavior.
- **Two duplicate LLM providers for the same class of work.** `ai-services/scripts/_api.py` already runs Groq (`llama-3.1-8b-instant`) as primary with Gemini as fallback; `backend/src/gemini.ts` was Gemini-only with no fallback at all. Rewrote `gemini.ts`'s LLM calls to match the Python side's provider order (Groq primary, Gemini fallback on missing/bad key or exhausted retries), so both halves of the system behave consistently and depend on one, cheaper primary provider instead of two. Also deleted `generateAIPersonalizedWorksheet` — exported and imported, but called by nothing anywhere in the codebase.
- **A 2.2M-line commit still sitting in the repo.** `Database/*.json` (commit `ceb6691`, "mongo database in json format") added ~49MB of unreferenced synthetic seed data — confirmed via grep that no code in `backend/`, `frontend/`, or `ai-services/` reads it, and confirmed the content itself is synthetic/generated (fake names, sequential fake ObjectIds, already-masked Aadhaar numbers), not a real PII leak. Removed it and gitignored the path so it can't silently recur.
- **Centralized the magic thresholds CLAUDE.md already flags as duplicated** (`MAX_LEVEL`, `CERTIFICATION_LEVEL`, `SCORE_BAND_STRONG`) into `shared/constants.ts`, backend-only for now. Also corrected a stale assumption along the way: `59` isn't the level cap (it's just curriculum level #59's own id) — `93` is, confirmed against `curriculumMap.ts`'s 93-node registry. Deliberately did *not* force-merge threshold values that only coincidentally share a numeral but aren't the same concept (e.g. a 3-tier 80/50 sub-level band vs. per-topic 70/60/50 concept-mastery bands) — left those as literals with a comment explaining why, rather than a band-aid unification that would misrepresent them as one shared value.

**One planned item was reverted after live testing showed it would break a real feature:** admin-gating `GET /api/icr/cloud-config` (to match its `POST` counterpart) turned out to be consumed by the actual scanning UI (`IcrTwoStageScan.tsx`) for every role, not just admins — it only exposes booleans ("is a provider configured?"), the sensitive write path (`POST`, which sets actual API keys) was already correctly admin-gated. Left as-is; noted in the PR for visibility rather than silently dropped.

## Test plan

No test framework existed in this monorepo before this PR (only `tsc --noEmit`). Added Node's built-in test runner (`node:test`, zero new dependencies) — `npm run test --workspace @fln/backend`, 15/15 passing:

- [x] `paperGenerator.test.ts` — `escapeHtml()` against all 5 HTML-significant characters, a benign name round-trips unchanged, an `onerror`-image injection attempt is neutralized
- [x] `gemini.test.ts` — Groq-primary/Gemini-fallback logic via a mocked `fetch`: Groq success never touches Gemini, Groq 401 is non-retryable and falls through, 429 retries exactly 3 times before giving up; deterministic grading still works when both providers are unavailable
- [x] `icr-rate-limit.test.ts` / `icr-size-cap.test.ts` — real HTTP requests against the app (via an exported `createApp()`, split out of `startServer()` specifically to make this possible) wrapped in an ephemeral server: confirms the rate limiter actually trips and both `evaluate-cloud` and `evaluate-pdf` (both of its code paths) actually reject oversized payloads. Writing this test caught a real ordering bug — the size check ran *after* the API-key lookup, so it was unreachable whenever no cloud key was configured; fixed.
- [x] `students-export.test.ts` — the `?all=1` limit math was pulled into an exported pure function (`resolveStudentsQueryLimit`) so the real 10,000-row ceiling is directly provable, instead of only provable by seeding 10,000+ rows into a test DB; plus an isolated-DB integration check that `all=1` still returns a full small dataset untruncated
- [x] `npm run lint` clean (`tsc --noEmit`, no new errors)
- [x] Manually verified against the real dev server: superadmin and teacher login both get `200` from `GET /api/icr/cloud-config` (confirming the revert above is correct); full live Puppeteer PDF generation with a malicious student name, no crash, no injected markup in the render


jgupta05072003-code:
Thank you for this contribution.

We have not yet done a full review of this pull request, since it currently has a merge conflict with the main branch. Kindly pull the latest main branch into your branch and resolve the conflict at your end.

Also, we could not find a linked issue in the pull request description. Kindly link the issue this pull request is meant to resolve, using "Closes #N", so that it can be tracked properly. If there is no existing issue for this work, kindly open one describing the motivation before we proceed with a review.

Once the conflict is resolved and the issue is linked, we will proceed with a full review.

PR #159, opened 2026-08-08
feat: add password reset flow, coordinator validation, fixed super-admin

sampurna222:
Expanded Coordinator Registration: SuperAdmin can now seamlessly create all system user roles—e.g.: Teachers, Principals, Volunteers, and Admins—directly from a single, unified registration interface.

Secure "Forgot Password" Flow: Implemented a complete password recovery solution. Users can now enter their registered email to receive a secure, automated reset link, allowing them to independently regain account access.


jgupta05072003-code:
Thank you, Sampurna. I reviewed this at `d3920436`. Password reset is one of the easiest things to get subtly wrong, and you got most of the hard parts right:

- **`randomBytes(32)` from `crypto`** — cryptographically secure, not `Math.random()`. Correct.
- **One-hour expiry**, checked on redemption at line 294.
- **Single use** — the token and expiry are cleared after a successful reset (lines 303–304).
- **Email enumeration handled deliberately.** Returning the same "If an account exists..." response for unknown emails, with a comment explaining why, is exactly right and is something many implementations miss.

Two things block the merge.

**1. The reset link is hardcoded to `localhost`.**

`backend/src/index.ts:182`:

```ts
const resetLink = `http://localhost:5173/reset-password/${resetToken}`;
```

In production every reset email would send the user to their own machine, so the feature cannot work at all once deployed. This needs to come from configuration — an `APP_BASE_URL` environment variable with the production origin, falling back to localhost for development.

**2. The reset token and the user's email are written to the server console in plaintext.**

Lines 183–186:

```ts
console.log(`🔑 PASSWORD RESET REQUESTED FOR: ${user.email}`);
console.log(`🔗 RESET LINK: ${resetLink}`);
```

Anyone who can read the server logs can take over any account by replaying a link, within the one-hour window. Logs get shipped to aggregators, tailed by people debugging unrelated problems, and retained far longer than an hour.

I understand why this is here — without SMTP configured locally there is no other way to get the link during development. That is fine, but gate it: only log when `NODE_ENV !== 'production'`, and log the token alone rather than the token next to the account it unlocks.

**Two more worth fixing, not blocking**

**3. The token is stored in the database in plaintext.** `updateUser(user.id, { resetToken, ... })`. If the users collection is ever read — a backup, a misconfigured query, the unauthenticated-endpoint class of bug we are fixing elsewhere — every live token is usable. Storing a SHA-256 hash and comparing hashes on redemption removes that. Small change, meaningful improvement.

**4. Both endpoints load every user into memory.** `dbStore.getUsers()` followed by `users.find(...)` at lines 167 and 294. That is a full scan of roughly 6,500 user records on every forgot-password and reset attempt, on endpoints that are unauthenticated and therefore easy to hammer. These should be indexed queries on `email` and on `resetToken`.

**Also**

The PR currently **conflicts with `main`** and will need a rebase.

`backend/src/index.ts` changes by **+2936/−6**, which is far more than a password reset flow. I can see geo data and coordinator registration in the file list too. Could you say briefly what else is in there? If this is several features in one PR, splitting them would let the password reset work merge on its own rather than waiting on review of everything.

---

The security thinking here is better than most of what I review — the enumeration handling in particular. Fix the two blockers and this is in good shape.

PR #162, opened 2026-08-15
Adding onboarding documentation

shrinidhisharma28:
Adding an onboarding document in markdown format after exploring the fln project repository and understanding how the system works, identified a gap in the current implementation and proposed a practical solution.


jgupta05072003-code:
Merged — thanks for the onboarding doc. No tracked issue is linked, which is expected for onboarding docs (no action needed here).

Issue #138, opened 2026-08-01
Move FLN_LEVELS_LIST + FLNLevelReferenceModal out of RoleDashboards.tsx


jgupta05072003-code:
## What this is for
`RoleDashboards.tsx` is a 3,100+ line file holding all 8 role dashboards. We're moving each piece into its own file so the codebase is easier to navigate.

## What to move
`FLN_LEVELS_LIST` (a data constant) and `FLNLevelReferenceModal` (a shared popup component that several dashboards use) — together, ~183 lines.

## Steps
1. Create `frontend/src/components/dashboards/FLNLevelReference.tsx`.
2. Cut both `FLN_LEVELS_LIST` and `FLNLevelReferenceModal` out of `RoleDashboards.tsx` into the new file, with whatever imports they need.
3. Back in `RoleDashboards.tsx`, add: `export { FLN_LEVELS_LIST, FLNLevelReferenceModal } from './dashboards/FLNLevelReference';` where they used to be.
4. Don't change any logic — copy it exactly as-is.

## Watch out for
- This one is used by other dashboards (search the file for `FLNLevelReferenceModal` and `FLN_LEVELS_LIST` to see which ones), so do this one **before** the dashboard-specific issues if possible — it'll make those easier.

## How to check your work
1. `cd frontend && npm run lint` — no errors.
2. `npm run build` — succeeds.
3. Run `npm run dev`, open the "FLN Level Reference" popup from any dashboard that has it, and confirm it still opens and shows the same content as before.

Part of #136.

jgupta05072003-code:
Saw PR #227 is already open for this — leaving it with you, not touching it myself.

Could you get it merge-ready by **end of day tomorrow (2026-08-20)**? If you hit a blocker, just comment here rather than letting it sit.

Issue #136, opened 2026-08-01
Epic: split RoleDashboards.tsx into one file per dashboard


jgupta05072003-code:
## Why
`frontend/src/components/RoleDashboards.tsx` is 3,122 lines — the single largest file in the whole repo. It holds all 8 role dashboards (Superadmin, Admin, School, Teacher, Volunteer, plus 2 shared pieces) in one file.

The good news: each dashboard is **already a fully independent, self-contained React component** — there's no shared state between them, and the whole file is only ever imported from one place (`App.tsx`). That makes this a much simpler split than the backend one (#116): mostly "move this component to its own file," not "figure out what to share."

## Approach
Move each `export const XyzDashboard` (and the two shared pieces) into its own file under `frontend/src/components/dashboards/`. Each sub-issue re-exports from `RoleDashboards.tsx` (`export { X } from './dashboards/X';`) so `App.tsx` never needs to change and the sub-issues don't conflict with each other. Do them one at a time, in separate small PRs.

## Sub-issues
- [x] #138 `FLN_LEVELS_LIST` + `FLNLevelReferenceModal` (~183 lines) — do this one early, other dashboards use it
- [x] #137 `SchoolDashboard` (~96 lines) — smallest, good first pick
- [x] #139 `RegionalAnalyticsView` (~292 lines)
- [x] #140 `AdminDashboard` (~405 lines)
- [x] #143 `SuperadminDashboard` (~694 lines) — biggest, do last
- [x] #142 `VolunteerDashboard` (~655 lines)
- [x] #141 `TeacherDashboard` (~791 lines) — most-used dashboard, be thorough testing it

## Suggested order
#138 (shared modal, unblocks others) → #137 (smallest) → #139 → #140 → #142 → #141 → #143 (biggest, last).

## Not in scope here
No behavior, layout, or logic changes — verbatim relocation only. If you spot an actual bug while doing this, file it separately instead of fixing it in the same PR.

jgupta05072003-code:
Confirmed complete as of 2026-08-22. All 7 sub-issues closed: #141, #142, #139, #143, #137, #140, #138 — the last four (#143, #137, #140, #138) finished between 08-20 and 08-21 by Vanshika and Nitish Naveen. `RoleDashboards.tsx` fully split into per-dashboard files under `frontend/src/components/dashboards/`. Closing.

PR #228, opened 2026-08-18
refactor: move AdminDashboard out of RoleDashboards.tsx

### LINKED ISSUE #140: Move AdminDashboard out of RoleDashboards.tsx
## What this is for
`RoleDashboards.tsx` is a 3,100+ line file holding all 8 role dashboards. We're moving each one into its own file so the codebase is easier to navigate.

## What to move
`export const AdminDashboard` (~405 lines) — the dashboard shown to State Admin-role users.

## Steps
1. Create `frontend/src/components/dashboards/AdminDashboard.tsx`.
2. Cut the `AdminDashboard` component out of `RoleDashboards.tsx` into the new file, with whatever imports it needs.
3. Back in `RoleDashboards.tsx`, add: `export { AdminDashboard } from './dashboards/AdminDashboard';` where it used to be.
4. Don't change any logic or JSX — copy it exactly as-is.

## How to check your work
1. `cd frontend && npm run lint` — no errors.
2. `npm run build` — succeeds.
3. Run `npm run dev`, log in as a State Admin user, and confirm the dashboard looks and behaves exactly the same as before.

Part of #136.

Osheun:
Moves `AdminDashboard` into `frontend/src/components/dashboards/AdminDashboard.tsx`, as described in the issue.

`RoleDashboards.tsx` goes from 3,138 to 2,738 lines.

### No logic changed

The 401 moved lines are byte-for-byte identical to the original — verified programmatically against `git show HEAD:...`, not by eye. Only imports were added on top.

### One decision worth your review

`AdminDashboard` depends on four things that live in `RoleDashboards.tsx` and were not exported. To keep the move to a pure cut-and-paste I exported them in place rather than relocating them:

| Symbol | Was | Now |
|---|---|---|
| `STATE_NAMES` | `const` | `export const` |
| `DISTRICT_NAMES` | `const` | `export const` |
| `DashboardProps` | `interface` | `export interface` |
| `RegionalAnalyticsView` | already exported | unchanged |

That does create an import cycle: `RoleDashboards` re-exports `AdminDashboard`, which imports back from `RoleDashboards`. It is safe as written — `DashboardProps` is pulled in with `import type` so it is erased at compile time, and the two `Record` constants are only read inside the component body at render, never at module evaluation, so there is no TDZ hazard. Rollup builds it without a circular-dependency warning.

Still, if the plan for #136 is to keep extracting dashboards, these shared pieces will keep being needed by each one. It may be cleaner to land a small `dashboards/shared.ts` holding `DashboardProps`, `STATE_NAMES` and `DISTRICT_NAMES` and break the cycle for good. I kept that out of this PR to respect "don't change any logic" — happy to do it here or in a follow-up, whichever you prefer.

### Checks

- `npm run lint` (`tsc --noEmit`) — clean
- `npm run build` — succeeds, no circular-dependency warning

I could not do the manual State Admin login check, which needs a seeded backend.

Independent of #138 — the two touch different regions of the file.

Closes #140


jgupta05072003-code:
Hi, thank you for the contribution here. We're closing this PR for now — we don't have an onboarding record for you on this project (the repo asks new contributors to add a short onboarding doc under `Ideas/` on their first PR), and without that we're not able to accept and merge external contributions right now. If you'd like to contribute going forward, please add an onboarding doc first and feel free to reopen or resubmit this work afterward. Sorry for the inconvenience, and thanks for understanding.

Issue #173, opened 2026-08-17
Rebuild Student List with class-wise subtabs


jgupta05072003-code:
## What's wrong

**Depends on the "Fix FLN-C" issue (replace hardcoded mock arrays with real data) being done first — otherwise this would just organize fake data more neatly.**

The entire `student_list` panel in `frontend/src/components/PanelViews.tsx` (lines 387–393) is a single `<EmptyStudents students={students} />` call — one flat, ungrouped list with no per-class organization. A teacher with multiple classes (2-A, 3-A, 4-A) has to scroll through everyone mixed together instead of looking at one class at a time.

## What to do

Rebuild the Student List page with class-wise subtabs — Class 2-A / 3-A / 4-A — instead of one flat list.

## How to do it

- File: `frontend/src/components/PanelViews.tsx` — rebuild the `student_list` panel with subtabs, filtering the existing `students` array by `classGroup`/`section`.
- Reuse the same tab pattern already used successfully on the Dashboard for its class tabs — don't invent a new pattern.
- No backend changes needed directly, but the data being displayed must come from the real student API, not the `STUDENTS_MOCK` fake array — this is why the "Fix FLN-C" issue has to land first.

## How to verify it's done

Open the Student List page as a teacher with multiple classes — students should be organized into separate class tabs, not one long mixed list.

PR #242, opened 2026-08-20
feat(certification): R-7 Certification Engine + onboarding doc

Sainikhil-hub:
## Summary
Implements SRS R-7 (Certification Engine) end-to-end:
- Backend: pure eligibility engine + MongoDB orchestrator with optimistic concurrency
- API: `POST /api/certification/review/:id` (SUPERADMIN/ADMIN), `GET /api/certifications`
- Auto-trigger: fires fire-and-forget at every `addEvaluationReport` site
- Frontend: `CertificationReviewPanel` (admin queue) + `CertificationHistoryCard` (per-student)
- Demo seed script: produces 2 review_needed + 1 active students for mentor demos
- Onboarding doc: `Ideas/ONBOARDING-Sainikhil.md` (per project policy)

## What changed (29 files in the feat commit,1 file in the docs commit)
- `backend/src/certification.ts` — pure `decideEligibility` engine (81 lines)
- `backend/src/certificationRecords.ts` — orchestration + optimistic concurrency (358 lines)
- `backend/src/competencyRequirements.ts` + `data/competencyRequirements.seed.json` — registry (168 lines)
- `backend/src/routes/certification.ts` — review + list endpoints (153 lines, NEW file)
- `backend/src/index.ts` — route registration (2 lines)
- `backend/src/routes/evaluation.ts` — 2 fire-and-forget triggers (7 lines)
- `backend/src/routes/students.ts` — 1 fire-and-forget trigger (4 lines)
- `backend/src/db.ts` — `Certification`, `CompetencyRequirement`, `MasteryLevel`, `CertificationStatus` types + 7 DBStore accessors (134 lines)
- `backend/src/scripts/demo-cert.ts` — demo seed (220 lines)
- `backend/src/modules/certification/` — future-Mongoose mirror + `notification.service.ts`
- `backend/src/__checks__/certification*.check.ts` — runnable assert scripts (zero deps)
- `frontend/src/components/CertificationReviewPanel.tsx` — admin queue (407 lines)
- `frontend/src/components/CertificationHistoryCard.tsx` — per-student history (167 lines)
- `frontend/src/components/Layout.tsx` — sidebar entry (4 lines)
- `frontend/src/components/PanelViews.tsx` — panel routing (14 lines)

## Test plan
1. `npm install && npm run lint` — clean
2. `npx tsx backend/src/scripts/demo-cert.ts --reset` → seeds 5 Class 4 students
3. Login as `superadmin@fln.org` / `Fln@2026`
4. Sidebar → Certification Reviews → see 2 review_needed rows
5. Click Confirm on Sushma Khan → toast + status flips to active
6. Click Revoke on Savita Ansari → modal asks reason → status flips to revoked
7. `curl` race test: two parallel review POSTs → one succeeds, other gets 409

## Onboarding
Per project policy: `Ideas/ONBOARDING-Sainikhil.md` includes all 6 required sections (What is FLN, FLN as a system, Current state, Gaps, Ideas, Your Contribution) and is ready for review.

## Verification of gaps (live data)
`/api/stats.certifiedCount` = 63,276 (uses `currentLevel >= 5` shortcut)
`/api/certifications?status=active` count = 2 (actual Certification rows)
→ documented as Gap 1 in the onboarding doc; the new engine doesn't fix this gap (out of scope for this PR) but the certification data is now the source of truth for the review workflow.

## Notes
- Cert rows live in MongoDB Atlas only — no data leaks via this PR (verified: 0 cert row matches in commit content)
- Demo-cert seed script is the canonical way for reviewers to get review_needed certs for their own demos


jgupta05072003-code:
Thank you for this, Sai Nikhil. This is a detailed and thoughtfully implemented contribution toward the R-7 Certification Engine.

A few things stood out positively during the review. The concurrency handling is not only mentioned in the PR description but is also properly implemented end to end. The optimistic concurrency flow, including the 409 conflict handling, is wired from the backend through to the frontend. The separation between the eligibility decision logic and the database/orchestration layer also makes the certification flow easier to understand, test, and maintain.

I also appreciate that the PR clearly mentions what is intentionally outside its current scope, particularly the existing currentLevel >= 5 statistics shortcut. Keeping that as a separate follow-up task makes the scope of this PR much clearer.

One point worth discussing further is the presence of two implementations of the eligibility engine: the live implementation in backend/src/certification.ts and the future Mongoose-oriented version under backend/src/modules/certification/. The manual validation script is a useful safeguard, but since we do not currently have CI configured to run it automatically, preventing drift still depends on someone remembering to run the check manually. We may want to decide whether this validation should be added to CI in the future, or whether the second implementation should wait until the Mongoose migration is actively being worked on.

Another important consideration is that the certification module is closely dependent on how the level progression logic is finalized and how student mastery is determined through the evaluation and report-generation flow. Since certification is ultimately based on the evidence produced by these systems, we should ensure that the certification criteria remain aligned with the final level logic, mastery calculations, and reporting methodology. Any future changes in these areas should be reflected consistently in the certification engine as well.

Sainikhil-hub:
Respected Ma'am,

Thank you very much for taking the time to review PR #242 so thoroughly.
Your feedback on both the concurrency handling and the engine's structure
is genuinely helpful, and I appreciate the detailed walkthrough.

I have noted both concerns you raised and would like to confirm my
follow-up plan:

1. *Two-engine drift risk (parallel Mongoose implementation):* Agreed —
this is a real concern. Since the Mongoose migration has not yet started,
the parallel implementation in backend/src/modules/certification/ is
currently dormant and adds maintenance surface area without any runtime
benefit. I will delete that entire directory, along with the related stub
files (backend/src/models/competency/, backend/src/interfaces/competency/,
and backend/src/utils/seedCompetencyRequirements.ts), in a follow-up commit
on the same PR. The live backend/src/certification.ts will then become the
single source of truth, and the Mongoose migration can re-introduce the
mirror from this stable reference when it actually begins.

2. *Coupling with level-progression and mastery-evaluation logic:* Agreed —
the certification engine's evidence base does sit on top of the level
generator, the evaluation report flow, and the conceptMastery calculation.
I will add a "See also" comment block at the top of
backend/src/certificationRecords.ts explicitly pointing to those modules,
so any future change in those areas prompts a corresponding cert-engine
review.

Both changes will be pushed as a single follow-up commit on PR #242, so the
final diff stays focused on R-7 without leaving dead code or undocumented
coupling.

Once the changes are in, I will share the updated PR link. Please let me
know if you would prefer a different approach for either item, or if there
is anything else you would like me to address.

Thank you again for your guidance and patience.

Sainikhil-hub:
Respected Ma'am,

Thank you very much for taking the time to review PR #242 so thoroughly.
Your feedback on both the concurrency handling and the engine's structure
is genuinely helpful, and I appreciate the detailed walkthrough.

I have noted both concerns you raised and would like to confirm my
follow-up plan:

1. *Two-engine drift risk (parallel Mongoose implementation):* Agreed —
this is a real concern. Since the Mongoose migration has not yet started,
the parallel implementation in backend/src/modules/certification/ is
currently dormant and adds maintenance surface area without any runtime
benefit. I will delete that entire directory, along with the related stub
files (backend/src/models/competency/, backend/src/interfaces/competency/,
and backend/src/utils/seedCompetencyRequirements.ts), in a follow-up commit
on the same PR. The live backend/src/certification.ts will then become the
single source of truth, and the Mongoose migration can re-introduce the
mirror from this stable reference when it actually begins.

2. *Coupling with level-progression and mastery-evaluation logic:* Agreed —
the certification engine's evidence base does sit on top of the level
generator, the evaluation report flow, and the conceptMastery calculation.
I will add a "See also" comment block at the top of
backend/src/certificationRecords.ts explicitly pointing to those modules,
so any future change in those areas prompts a corresponding cert-engine
review.

Both changes will be pushed as a single follow-up commit on PR #242, so the
final diff stays focused on R-7 without leaving dead code or undocumented
coupling.

Once the changes are in, I will share the updated PR link. Please let me
know if you would prefer a different approach for either item, or if there
is anything else you would like me to address.

Thank you again for your guidance and patience.

jgupta05072003-code:
For now, `backend/src/` will continue to follow the existing flat structure. New backend work should go into `backend/src/routes/<domain>.ts`, exporting `register<Domain>Routes(app)`, with reusable business logic placed in plain modules under `backend/src/services/`.

At this stage, we will not introduce additional top-level structures such as `modules/`, `controllers/`, `repositories/`, `models/`, or `interfaces/`.

This applies equally to #242 and #391. Both PRs independently raised the same architectural question, so it seems better to document one repository-wide decision rather than address it differently on each PR.

Neither implementation is inherently wrong. Both are reasonable architectural approaches. The main concern is that the repository did not previously have a clear decision on this, and that ambiguity should have been resolved earlier.

## Why

**1. The two proposed structures follow different conventions.**

#242 is primarily domain-scoped:

```
modules/certification/services/...
```

while #391 is primarily layer-scoped:

```
controllers/
models/
interfaces/
```

Both approaches are valid in isolation. However, merging both alongside the existing `routes/` structure would leave the backend with three different organisational conventions.

That would make it increasingly difficult for future contributors to know which pattern they should follow.

For now, keeping one convention across the backend seems preferable.

**2. Both PRs introduce an additional database-access approach.**

The current backend uses the native `mongodb` driver through `db.ts` and its `dbStore` accessors.

Both PRs also introduce Mongoose schemas.

Using both approaches is technically possible, but doing so without an explicit repository-wide decision could leave us with two different ways to access the same data and no clear guidance on:

- which layer is authoritative;
- where new data-access logic should be added;
- how transaction behaviour should remain consistent;
- and how both representations stay aligned over time.

There is also a dependency-management point worth noting: `backend/package.json` currently declares `mongodb`, while `mongoose` is declared in the root `package.json`. The current imports therefore rely on workspace dependency resolution rather than a backend-local declaration.

This is better treated as an explicit architectural decision rather than something introduced gradually through individual feature PRs.

**3. The strongest benefit of additional repository layers is not fully available yet.**

One of the strongest arguments for repository and service abstraction is the ability to isolate and test those layers independently.

At present, the backend has very limited automated test coverage.

Because of that, we would be introducing the structural complexity immediately without yet receiving the full testing benefit that normally helps justify it.

Once the backend test suite becomes stronger, this decision can reasonably be revisited.

**4. Adopting a new architecture properly would require a wider migration.**

The backend already contains 21 existing route files following the current convention.

If we decide to adopt a layered structure, it would be better to do so as a deliberate repository-wide migration rather than applying it only to selected new features.

Without such a migration plan, both structures are likely to remain in parallel for a long time.

## What this asks of each PR

### #242 — R-7 Certification Engine

Please keep the certification feature, but align its structure with the current backend convention.

The preferred direction would be:

```
backend/src/routes/certification.ts
backend/src/services/certification*.ts
```

Database reads and writes should go through the existing `dbStore` mechanism rather than introducing a separate Mongoose data-access path.

The current `certification.model.ts` is not yet part of the running backend, so this should be a good point to consolidate the implementation before other code begins depending on it.

It would also be helpful to keep the certification implementation in one clear location rather than maintaining parallel versions that could gradually diverge.

`AGENTS.md` appears unrelated to the certification feature itself and would be better handled separately.

Most importantly, the certification functionality itself is absolutely valuable and wanted. This request is only about bringing the implementation into alignment with the repository's current backend structure.

### #391 — Remediation

The same structural principle applies here.

The preferred direction would be:

```
controllers/*.controller.ts  →  routes/*.ts
```

Mongoose-backed data access should be adapted to the existing `dbStore` approach, and shared TypeScript types can generally remain close to the code that uses them rather than introducing another top-level `interfaces/` convention.

The PR is also currently conflicting, so it will need to be rebased against the latest target branch.

Separately, it would be helpful to remove development/debug artifacts that do not need to be part of the final implementation, including:

```
db_diff*.txt
icr_diff.txt
backend/diagnostic_output.json
```

The empty `RemediationGenerationCard.tsx` can also be removed if it is not currently being used.

Please also correct:

```
remediaitionFormatter.ts
```

to:

```
remediationFormatter.ts
```

## What the intended structure looks like

`vicharanashala/tenali#173` is a useful reference specifically for backend structure.

That PR breaks a very large `server/index.js` into grouped route files and adds an API contract test, while keeping the architecture relatively simple and avoiding additional backend layers or another database abstraction.

That is broadly the kind of modularisation we would like to follow here.

The PR may have other implementation concerns being discussed separately; the reference here is only to its structural approach.

## When this decision should be revisited

This is not intended to be a permanent rejection of layered architecture.

There are at least two concrete situations where the decision should be reopened.

**1. A meaningful backend test suite exists.**

Once there is sufficient automated test coverage, the isolation benefits of repository and service abstractions become much more useful and measurable.

At that stage, a layered architecture would be worth evaluating again.

**2. Individual route files become difficult to maintain.**

If a route file grows substantially — for example, to roughly 800 lines or beyond — that would be a reasonable indication that the current structure is beginning to strain.

At that point, additional domain-level or architectural separation may be justified.

A preference such as "this feels cleaner" alone may not justify the migration cost, but measurable maintainability or testing needs certainly can.

If either situation arises, please open an issue referencing this ADR so that the architecture can be reconsidered with the full context available.

## Not decided here

This decision does not determine whether Mongoose should eventually replace the native MongoDB driver across the repository.

That is a separate architectural question and deserves its own discussion based on migration cost, testing, maintainability, transaction handling, and long-term requirements.

The decision here is simply that we should avoid gradually maintaining two database-access approaches and multiple backend structures unless we have explicitly chosen to do so repository-wide.

## Where to find this decision

Recorded in the repository at:

```
docs/adr/001-backend-structure.md
```

It is being added in **#427** — https://github.com/vicharanashala/fln/pull/427

Once that merges, the file will be readable on `main` at:

https://github.com/vicharanashala/fln/blob/main/docs/adr/001-backend-structure.md

Future architecture discussions can reference it directly, and any request to revisit the decision should link to it so the original reasoning is available.

Sainikhil-hub:
Respected Mam,

Thank you for the detailed review, and for capturing this as a repo-wide
ADR rather than resolving it differently per PR — I agree one documented
convention is the right call, and the reasoning (avoiding three parallel
structures, keeping one authoritative data-access path, and not adding
abstraction layers before tests can benefit from them) makes sense to me.

Here's what I'll change in #242:

Align with the current convention. The wired-in implementation already
follows it — routes/certification.ts exports
registerCertificationRoutes(app) and all reads/writes go through the
existing dbStore. The modules/certification/ tree is an earlier draft that
survived the merge with upstream's routes-split; nothing in the running
backend imports it (verified against routes/, index.ts, db.ts and the check
scripts). I'll delete it outright, including certification.model.ts and the
Mongoose repository, so there's exactly one implementation and no second
data-access path in this feature.
Business logic into services/. The three logic files move to
backend/src/services/ (certificationEligibility.ts,
certificationRecords.ts, certificationCompetencyRequirements.ts), keeping
routes/certification.ts as the only HTTP layer. The PR introduces no
Mongoose usage.
AGENTS.md. Agreed it's unrelated — I'll remove it from this PR and raise it
separately. When I do, I'll also revise its backend-structure guidance to
match the ADR, since it currently points new domain work at
modules/<domain>/, which this decision supersedes.
I'll rebase onto latest main, re-run npm run lint and the __checks__
scripts, and re-verify the review/queue endpoints end-to-end before pushing.
One question: Ideas/ONBOARDING-Sainikhil.md is also in this PR — happy to
keep it here or move it to its own PR if you prefer; either way I'll update
it to describe the final structure.

Thanks again — the reference to #173's structural approach is helpful, and
I'll cite docs/adr/001-backend-structure.md in future architecture
discussions.

Sainikhil-hub:
Respected Mam,

Thank you for the detailed review, and for capturing this as a repo-wide
ADR rather than resolving it differently per PR — I agree one documented
convention is the right call, and the reasoning (avoiding three parallel
structures, keeping one authoritative data-access path, and not adding
abstraction layers before tests can benefit from them) makes sense to me.

Here's what I'll change in #242:

Align with the current convention. The wired-in implementation already
follows it — routes/certification.ts exports
registerCertificationRoutes(app) and all reads/writes go through the
existing dbStore. The modules/certification/ tree is an earlier draft that
survived the merge with upstream's routes-split; nothing in the running
backend imports it (verified against routes/, index.ts, db.ts and the check
scripts). I'll delete it outright, including certification.model.ts and the
Mongoose repository, so there's exactly one implementation and no second
data-access path in this feature.

Business logic into services/. The three logic files move to
backend/src/services/ (certificationEligibility.ts,
certificationRecords.ts, certificationCompetencyRequirements.ts), keeping
routes/certification.ts as the only HTTP layer. The PR introduces no
Mongoose usage.
AGENTS.md. Agreed it's unrelated — I'll remove it from this PR and raise it
separately. When I do, I'll also revise its backend-structure guidance to
match the ADR, since it currently points new domain work at
modules/<domain>/, which this decision supersedes.

I'll rebase onto latest main, re-run npm run lint and the __checks__
scripts, and re-verify the review/queue endpoints end-to-end before pushing.
One question: Ideas/ONBOARDING-Sainikhil.md is also in this PR — happy to
keep it here or move it to its own PR if you prefer; either way I'll update
it to describe the final structure.

Thanks again — the reference to #173's structural approach is helpful, and
I'll cite docs/adr/001-backend-structure.md in future architecture
discussions.

jgupta05072003-code:
Verified the follow-up: backend/src/modules/certification/, models/competency/, and interfaces/competency/ are gone from this branch — backend/src/certification.ts (via certificationRecords.ts) is the single source of truth now, no drift risk. One small thing not fully done: the promised 'See also' comment pointing certificationRecords.ts at the level-generator/evaluation/mastery modules isn't there yet — but that's a documentation nice-to-have, not a blocker. Concurrency handling and eligibility/orchestration separation both check out. This is ready to merge.

PR #259, opened 2026-08-21
refactor(frontend): move SchoolDashboard out of RoleDashboards.tsx (#…

vanshikaagrawalcsaiml24:
…137) i have solved this issue
Screenshot 2026-08-21 112749


jgupta05072003-code:
Thanks Vanshika — the SchoolDashboard extraction itself looks good, I tested it live (build clean, dashboard renders, worksheet flow works, no console errors).

One thing to fix before this is ready to merge: `frontend/vite.config.ts` has an unrelated change — the `/api` proxy default port changed from 3000 to 5000, but `/output` and `/worksheets` are still pointing at 3000. Looks like a local dev-config change that got committed by accident. Could you revert that one line (or point all three at the same port if it was intentional) and push again?

PR #114, opened 2026-08-01
feat: secure Aadhaar tokenization and step-up detokenization

DUKartik:
# feat(aadhaar-vault): rebuild Aadhaar protection in-process (closes #406)

## Objective

Rebuild the Aadhaar tokenization + Step-Up detokenization protections from
PR #114 **inside the existing FLN Express + MongoDB backend**, without adding
a new microservice, a new database engine, or new deployment surface.

Closes #406.

Companion / follow-up to PR #114 ("feat: secure Aadhaar tokenization and
step-up detokenization"). That PR's security engineering is preserved verbatim
(envelope encryption, server-derived challenge targets, single-use challenges,
audit-before-state-transition, plaintext zeroing). The architecture is
changed: it now lives as a self-contained module inside the backend instead
of a separate process.

### 1. Plaintext Aadhaar no longer exists at rest in the FLN database

`Student.aadharMasked` previously held the raw 12-digit Aadhaar for
superadmin (and was the plaintext field for everyone else too — only the
on-wire projection was masked). It now holds a deterministic identity id
(`aadhaarIdentityId`) plus a per-record opaque vault token
(`aadhaarTokenId`) and the display mask (`aadharMasked: "XXXX-XXXX-1234"`).

The actual ciphertext is held by the in-process vault module under
`backend/src/modules/vault/`, encrypted with AES-256-GCM under a per-record
DEK wrapped by HKDF-SHA-256(master, salt=context, info="aadhaar-vault/dek-wrap").

- A read-only at-rest audit (`backend/scripts/audit-aadhaar-at-rest.ts`,
  `npm run audit:aadhaar`) asserts no plaintext 12-digit Aadhaar appears
  in the FLN MongoDB collections. The suite still passes after every change
  in this PR.

### 2. Step-Up auth (TOTP + server-derived challenge + single-use) is required to reveal

Revealing the full Aadhaar now requires:

1. **Role gate** — `canAccessStudent` blocks non-admins at the FLN layer.
2. **TOTP enrollment** — admin enrolls a TOTP factor once via
   `POST /api/students/:id/aadhaar/mfa/enroll` (RFC 6238 via `otpauth`,
   8 digits, 30 s period, ±3 windows = ±90 s acceptance).
3. **Server-derived step-up challenge** — `POST /api/students/:id/aadhaar/step-up/request`
   mints a challenge. The vault token id is resolved **from the authorized
   student record**, not from the request body — a caller cannot supply an
   arbitrary token to bypass authorization (security rule #7).
[...]

PR #285, opened 2026-08-22
doc(onboarding): ONBOARDING-Kartik-Nath-Khatri.md

DUKartik:
## Onboarding Document

Added an onboarding document describing the FLN platform, the identified Aadhaar security gap, and the proposed Aadhaar Vault architecture.

It explains the motivation, design, tokenization and Step-Up detokenization flows, security model, and implementation details of **PR #114** for future contributors and reviewers.

**Reference:** [PR #114](#114)

### DIFF
--- a/Ideas/ONBOARDING-Kartik-Nath-Khatri.md
+++ b/Ideas/ONBOARDING-Kartik-Nath-Khatri.md
@@ -1,0 +1,912 @@
+# ONBOARDING — Kartik Nath Khatri
+
+## 1. What is FLN?
+
+FLN stands for **Foundational Literacy and Numeracy** — the ability to read with comprehension and perform basic arithmetic, including letter and word recognition, reading fluency, basic comprehension, number sense, and simple operations. It broadly covers the foundational years from pre-school through Grade 3.
+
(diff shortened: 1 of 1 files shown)


jgupta05072003-code:
Thank you for this, Kartik. This is a strong and well-grounded onboarding document.

I also verified a few of the more important architectural claims directly against the implementation referenced in the document, including the fail-closed registration behavior, AES-256-GCM encryption, HKDF-SHA-256 key derivation, and the stated test-suite count. These checks matched what the document describes, which gives confidence that the write-up is reflecting real implementation rather than presenting an aspirational design.

What I particularly appreciate is the transparency in Sections 18 and 19. The document explicitly mentions the known test failures and clearly separates production-ready functionality from the parts that are still development-only or incomplete. For a security-sensitive contribution, that distinction is important.

One small documentation suggestion: Section 16, "Security Properties," currently reads as a list of implemented guarantees. It may help to add a short forward reference there to Section 19 so that readers who skim the document still immediately see the current limitations and development-only components.

Overall, this is a thorough and trustworthy onboarding document. Nice work, and thank you for the care taken in documenting both the implementation and its current limitations.

PR #306, opened 2026-08-22
Feat : adaptive micro practice engine

Ashishpatra0-0:
## What this PR does

Adds **Micro-Practice** -  a spaced-repetition practice system that generates small, targeted printable papers for a student's specific weak competencies, tracks real difficulty progression as they improve, and provides a full workflow for teachers to evaluate completed papers.

## What's included

- **Generation** -  reuses the existing question engine to build small (3–5 question), single- or multi-competency papers targeted at a student's actual weak areas. Class-scoped bulk generation for a whole class at once.
- **Real difficulty progression** -  good performance genuinely advances a student through harder curriculum variations and levels, not just a disconnected counter. A competency is marked mastered only once genuinely out of content for it.
- **Re-assessment reconciliation** -  when a student is re-diagnosed, their existing practice progress is reconciled: resolved if no longer weak, reset if their placement changed significantly, left untouched otherwise.
- **Evaluation**  - QR-based paper identification, photo/PDF upload (including multi-page support), manual answer entry with real question context, and a new "visual-confirm" answer type for previously-ungradeable drawing questions.
- **Bulk evaluation** -  upload and grade many completed papers at once, grouped and navigable as a batch.
- **Scheduling** -  history-aware spaced repetition: intervals now factor in recent performance consistency (not just the latest score) before deciding how much to lengthen or shorten the next practice gap, independent per competency.


jgupta05072003-code:
Thank you for this, Ashish. I spent some time tracing the core logic in detail, including the schedule reconciliation, the spaced-repetition interval calculation, and the duplicate-question exclusion handling, rather than relying only on the PR description. The underlying design holds up well, and the problem this feature is trying to solve is an important one for the platform: turning a known weakness such as Fractions into repeated, targeted practice rather than waiting for the next broad diagnostic.

There are two relatively contained changes I would request before merge.

1. **Please align the on-screen manual practice flow with the PDF-generation flow.**

At the moment, `/api/practice/generate/:studentId` uses `generateQuestionsForLevel()`, which in this level range falls back to the placeholder generation path in `conceptQuestionGenerator.ts`. This is also the behavior that issue #359 is already trying to move away from for production use.

The manual flow also appears to search around `student.currentLevel ± 5` rather than using the student's exact Micro-Practice schedule position.

The PDF-generation flow already seems to handle both of these concerns correctly through `generateMicroPracticePaper()`, which uses the real generators from `levels_main.html` and respects the scheduled `levelId` and `subIdx`. Since that function already returns the generated questions as structured data in addition to producing the PDF, could we please reuse the same path for the on-screen flow as well?

That would help ensure that both ways of delivering Micro-Practice use the same content source and the same progression logic.

2. **Please move the new Micro-Practice routes out of `backend/src/index.ts`.**

The PR currently adds the `/api/practice/*` and `/api/students/:id/micro-practice/*` endpoints directly into `index.ts`. Recent backend work has been moving feature-specific routes out of this file and into `backend/src/routes/*`, so it would be helpful for this feature to follow the same structure.

A dedicated file such as `backend/src/routes/microPractice.ts` would keep the feature easier to maintain.

I would also suggest moving `calculateNextScheduleState()` out of the `startServer()` closure and into the relevant service layer, likely alongside the other scheduling logic in `practiceScheduleService.ts`. Since this function contains the core progression and interval logic, keeping it exported and isolated would also make it much easier to unit-test directly.

I do not think either of these changes requires expanding this PR into the larger question-bank or misconception-mapping work already planned under the later milestones. Those are broader pieces of work and can remain separate. The requests here are mainly about making the two Micro-Practice delivery paths internally consistent and keeping the new feature aligned with the backend structure the repository is already moving toward.

Overall, the scheduling and remediation flow here is thoughtfully designed, and I appreciate the amount of work that has gone into it. These changes would make the feature much easier to maintain and give us a cleaner foundation for the deeper concept-mastery and misconception work that follows.

Ashishpatra0-0:
@jgupta05072003-code Thank you for the detailed review, both requested changes are now complete:

1. On-screen practice flow now reuses the real generation logic. /api/practice/generate/:studentId now calls generateMicroPracticePaper() directly, resolving the student's real Micro-Practice schedule position (levelId/subIdx) instead of the old currentLevel ± 5 search and placeholder generator. It also now persists a MicroPracticePaper record, matching the PDF flow, so the returned paperId is usable via the existing /api/practice/paper/:paperId and grading endpoints. I confirmed this with a real test: it correctly resolved a student's actual schedule position, ran the real generator, and returned the same question shape as the printed papers. 
Note : this route currently has no frontend caller, so I verified it via direct API calls rather than the UI.

2. Micro-Practice routes moved into their own file. All 15 routes now live in backend/src/routes/microPractice.ts, following the same pattern as routes/students.ts. calculateNextScheduleState() (and its helper) moved into practiceScheduleService.ts and is now exported for direct unit testing, alongside the other scheduling logic there. I also re-verified end-to-end afterward -> generation, upload, partial/full grading, Due Today, and a real diagnostic submission all still work correctly, including confirming routes/students.ts's existing call into the shared service module was unaffected by the move.

Also merged in the latest main and resolved the resulting conflicts. Let me know if you'd like anything adjusted further.

Thank you again for the thorough review!

jgupta05072003-code:
Thank you for this work, and apologies for the delay in coming back to you. We have now reviewed the branch as it stands today.

**1. The two earlier points are addressed**

We can see that the sixteen routes have been moved out of `index.ts` into `backend/src/routes/microPractice.ts`, that `calculateNextScheduleState()` now lives in `practiceScheduleService.ts`, and that `/api/practice/generate/:studentId` calls `generateMicroPracticePaper()`, so the on-screen flow and the PDF flow now serve the same content. Thank you for making those changes.

**2. The level data source has changed since you opened this PR**

This one is not something you could have known, and it is not a fault in your work. `backend/src/flnLevels.ts` in this branch is a newly hand-authored map of levels to strands in the 59 level number space. We have since decided to retire the 59 space and work only in the 93 level space.

The background is that six separate copies of the level list had built up on `main`, because the 93 levels existed only as TypeScript in `frontend/src/data/skillProgressionMap.ts`, which the backend cannot import. So every feature that needed curriculum data wrote its own copy. `flnLevels.ts` would be the seventh, in the space we are removing.

**3. What we would suggest using instead**

`curriculumLevels` is now a real collection, with accessors on `dbStore`: `getCurriculumLevel`, `getCurriculumLevels`, `getCurriculumLevelByLegacy59` and `getCurriculumCoverage`. It would be helpful if the three import sites (`index.ts`, `routes/microPractice.ts` and the service) read from those instead of from `flnLevels.ts`. `getCurriculumLevelByLegacy59` should cover the cases where you currently look up by a 59 space number.

We would like to be clear that the micro practice work itself is wanted. The interval calculations and the schedule reconciliation both read as carefully done. This is only about where the level data comes from.

**4. Two smaller items before merge**

It may be preferable to regenerate `package-lock.json` from `main` rather than shipping the one in this branch. The change is currently +2791 and -5238 lines, and a plain `npm install` has previously moved Puppeteer's pinned Chrome version past what the server had cached, which took paper generation down in production.

The branch also currently conflicts with `main`, so a rebase would be needed before it can merge.

Please let us know if any part of the `curriculumLevels` change is unclear, and we can go through it together.

Ashishpatra0-0:
@jgupta05072003-code Before starting the `flnLevels.ts` → `curriculumLevels migration`, I investigated `getCurriculumLevelByLegacy59()`'s actual current data, and found the bridge isn't populated yet — every level's `legacyLevel59` field seeds as `null`, since the reviewed crosswalk file (`fln_59_to_93_crosswalk.json`) doesn't exist yet, only a `.PROPOSED` version explicitly marked unusable in its own commit message.

I also found that `feat/question-review-mapping` has moved away from the legacy-to-93 crosswalk approach entirely for levels 22-59 (where 8 of our 9 competencies live), using per-question tagging instead — a different mechanism than a level-to-level lookup.

Given this, migrating right now would mean `getSubsCountForLevel`/`getNextLevelInStrand` silently resolve to nothing, causing a correctness regression (defaulting to 1 sub-level for everything) rather than a visible error. Since Micro-Practice's `currentLevelId` never touches the 93-scale directly today (confirmed safe, no live bug), would it make sense to hold off until either the reviewed crosswalk lands, or there's clarity on whether legacy-to-93 lookup is still the intended path for levels 22-59 versus the per-question tagging approach? Happy to discuss further or take a different approach if you'd prefer.

jgupta05072003-code:
@Ashishpatra0-0 — **you are right, and thank you for checking rather than just doing what I asked.**

I verified every claim in your comment:

- The only crosswalk artefacts in the repo are `Research/fln_59_to_93_crosswalk.PROPOSED.md` and `.PROPOSED.json`. There is no reviewed `fln_59_to_93_crosswalk.json`.
- `seedCurriculumLevels.ts:158` says so in its own log line: *"legacyLevel59: null and no content flags. Re-run once the reviewed crosswalk lands."*
- `getCurriculumLevelByLegacy59()` exists at `db.ts:2497`, but with `legacyLevel59` seeding as `null` for every level it would resolve to nothing.

So your conclusion is correct: migrating now would make `getSubsCountForLevel` and `getNextLevelInStrand` silently default to 1 sub-level for everything — a quiet correctness regression rather than a visible failure. That is worse than not migrating.

**Decision: do not migrate in this PR.** Keep `flnLevels.ts` as-is. The migration is blocked on the pedagogy crosswalk review, which is on us, not on you. I will open a separate issue for it that links back to this comment so the reasoning is not lost, and it can be picked up once the reviewed crosswalk exists.

This is exactly the right way to handle a review request you think is wrong. Investigating and coming back with evidence saved us a regression.

---

**On the PR itself.** I checked out `dcba98d8` and reviewed it.

**What is good:**
- Both `tsc --noEmit` runs pass clean. The two new dependencies (`jsqr`, `pdfjs-dist`) are properly declared in `frontend/package.json` — I only saw errors before installing them, which was my environment, not your PR.
- The `Micro-Practice` nav entry is added in both role branches correctly, not duplicated.

**Three things to address:**

**1. `package-lock.json` churns by +2791/−5238.**

565 dependency-resolution lines change, with no `lockfileVersion` bump. That usually means the lockfile was regenerated with a different npm version than the rest of the team uses. A lockfile diff five times the size of your feature makes the PR hard to review and will conflict with every other open PR.

Please restore the lockfile from `main` and re-add only your two new dependencies, so the diff shows just those.

**2. `frontend/public/worksheets/levels_main.html` changes by +517/−204.**

I could not work out why a micro-practice PR rewrites a worksheet asset. If it is required, please say what changed and why. If it is incidental, please revert it.

**3. `Layout.tsx` is becoming a contended file — this one is not your fault.**

Five open PRs now modify `frontend/src/components/Layout.tsx`: this one, #418, #308, #242 and #393. Your actual change is two `list.push` lines, but the diff is +91/−95 because the surrounding block got reformatted, which will conflict with all four of the others.

Please keep the two nav lines and revert the reformatting. I will handle sequencing the rest.

---

Fix 1 and 3, answer 2, and this is close. Sorry for the wait on this one — you have been following up since 22 August.

jgupta05072003-code:
**A design question that has just become important — please read before doing more work on this PR.**

We are standardising extraction on **Ollama Gemma 4** and retiring the client-side ICR/OCR scanning path. Gemma 4 is already the single OCR provider in the backend (`backend/src/routes/evaluation.ts:19`; line 59 rejects anything else).

Looking at your two new components against that decision:

**`MicroPracticePaperUpload.tsx` — keep this.** Reading the QR code with `jsQR` to identify *which* paper and student a photo belongs to is paper identification, not answer extraction. It is orthogonal to Gemma and genuinely useful. Your fallback message when no QR is found is a nice touch.

**`MicroPracticeAnswerEntry.tsx` — this is the one to reconsider.** It renders PDF pages to canvas with `pdfjs-dist` so a teacher can look at the paper and type the answers in by hand. That is 500 lines building a manual path for exactly the job Gemma 4 is meant to do automatically.

The direction we are going is: photograph the sheet → Gemma extracts the answers → the teacher confirms or corrects. Not: teacher reads the paper on screen and types every answer.

So before you invest further, a question rather than an instruction:

**Should micro-practice answer entry call the Gemma extraction path instead of being manual?** If yes, the screen becomes a much smaller confirm-and-correct view over Gemma's output, and `pdfjs-dist` may not be needed at all. If there is a reason micro-practice papers cannot go through Gemma — a different layout, no QR, a size limit — tell me and the manual path is justified.

I would rather ask now than have you polish 500 lines we then remove.

**This does not change my earlier three points** — the `package-lock.json` churn, the `levels_main.html` change, and reverting the `Layout.tsx` reformatting all still stand, and are worth doing regardless of how this question resolves.

And to be clear, none of this is a criticism of the work. The decision changed after you built it.

jgupta05072003-code:
Recorded this as **#433** — *Retire the client-side ICR/OCR scanning path; standardise extraction on Ollama Gemma 4*. The micro-practice answer path question is work item 5 there, so please reply on the issue rather than here if that is easier.

**One thing I found while writing it up that affects this PR directly.**

`MicroPracticePaperUpload.tsx:138` calls:

```ts
const res = await apiFetch('/api/icr/rasterize-pdf', { ... })
```

**That endpoint does not exist.** I searched `backend/src/` — the only ICR routes are `/api/icr/evaluate-cloud`, `/api/icr/evaluate-bulk` and `/api/icr/cloud-config`, all in `routes/evaluation.ts`. There is no `rasterize-pdf` route anywhere.

So the PDF upload path cannot work as written; the request will fail. Either the endpoint needs implementing (`ai-services/scripts/pdf_rasterize.py` exists, so the rasterisation itself is available — it just is not exposed over HTTP), or the call needs removing.

Worth checking whether you tested this with an image rather than a PDF, since the image path does not hit that endpoint.

This is listed as work item 6 on #433. Your other items — the lockfile churn, `levels_main.html`, and the `Layout.tsx` reformatting — are unchanged.

Ashishpatra0-0:
@jgupta05072003-code I've addressed all three items:

1. `package-lock.json` — restored to match `main` exactly, then re-added only our two real dependencies (`jsqr`, `pdfjs-dist`). Diff is now +270/0, entirely their own dependency trees — down from +2791/-5238.

2. `levels_main.html` — this is deliberate, not incidental. It comes from 6 commits made throughout development, all directly needed for Micro-Practice:
 - Visible A/B/C match-question labeling across ~31 generators, so a teacher can grade by hand without tracing lines visually.
 - Question numbering added to vertical arithmetic questions (previously missing).
 - Duplicate/hanging-question fixes in several shared generators — including a genuine infinite loop in one generator when asked for more questions than its content pool could mathematically support.
 - `generateMicroSet()` — our actual generation entry point, entirely new/additive, doesn't modify the general worksheet path.
 - Drawing-question "visual-confirm" support for 5 previously-ungradeable question types (fraction shading, symmetry, division pictures), each with its own reference-image renderer.
 - A fix for composite generators that were silently ignoring the requested question count.

None of this touches the general worksheet pipeline's answer-key builders (`buildCleanAnswerKey`/`buildQuestionPaperJSON`), and every shared-generator change is backward-compatible — output shape and behavior for the general/diagnostic worksheet path is unchanged. Happy to walk through any specific piece in more detail.

3. `Layout.tsx` — restored to match `main`, with only the 3 real lines re-applied (the `Repeat` icon import and the two Micro-Practice nav entries, Teacher and Volunteer). Diff is now +3/-1.

On issue #433, work item 6 (`/api/icr/rasterize-pdf`): I checked this directly — the route does exist on this branch, at `index.ts:160`. It wasn't moved into `routes/microPractice.ts` during the earlier route-organization refactor since it's technically an ICR utility route, even though only Micro-Practice currently calls it — easy to miss if checking `routes/microPractice.ts` directly, since that's where every other Micro-Practice route now lives. It also correctly doesn't exist on `main`, since it was introduced entirely on this branch, which may be where the "does not exist" read came from if compared there instead. Happy to move it into `routes/microPractice.ts` for consistency if useful.

On issue #433, work item 5 (Gemma OCR for micro-practice answers): For typed/choice-style answers, I think switching to the Gemma extract-then-confirm flow makes sense and would reduce teacher effort, consistent with where the platform is heading. For the "visual-confirm" drawing questions specifically (tally marks, fraction shading, symmetry, division pictures), I'd suggest keeping the current reference-image-comparison approach — these aren't text-extraction problems, since there's no discrete value to extract, only a visual match to confirm. Routing them through Gemma would mean either asking it to interpret freehand drawings as text (unlikely to be reliable), or building a separate visual-comparison capability that's effectively what's already built, without the added AI dependency. `pdfjs-dist` would still be needed for rendering pages for these visual-confirm comparisons even if typed-answer entry moves to Gemma, so it likely isn't fully removable. Happy to discuss further if you see it differently.

Let me know if anything needs adjusting.

jgupta05072003-code:
Thank you for this, and apologies that it has been open a while. An adaptive micro-practice engine is close to the heart of what FLN is for — diagnosing a gap is only half the job, and practice targeted at that gap is the other half.

At 17 files and +5803/−237 — including `paperGenerator.ts` (+456), a 954-line route file, a 1739-line component and +226 to `db.ts` — this needs a dedicated review pass rather than a batch merge. It is queued for that rather than blocked.

Three things that would help when we get to it:

1. **`backend/src/flnLevels.ts` (new, 135 lines).** There is existing level-mapping machinery in the repo — `backend/src/config/curriculumMap.ts` and the `skillLevelMap` snapshot — and level identifiers are a known sore point here. Could you say in the description how `flnLevels.ts` relates to those: does it wrap them, replace them, or introduce a third mapping? If it is a third, that is the thing we will want to resolve before merge, because divergent level identifiers cause quiet, hard-to-trace wrongness later.

2. **`frontend/public/worksheets/levels_main.html` (+517/−204).** A large change to a generated-worksheet template is worth calling out separately, since it affects what actually gets printed and put in front of a child.

3. **How practice items are selected.** This is the pedagogical core. A short description of the selection rule — and in particular whether it is deterministic or model-generated — would frame the whole review. If it is model-generated, we will want to know what stops an unsuitable question reaching a child.

**One repo-wide note that may save you rework.** A decision was recorded in #288 that mastery is server-authoritative and that we do not want a second mastery model in the codebase; #289 covers how that gets wired. If your scheduling logic in `services/practiceScheduleService.ts` maintains its own notion of what a child has mastered, it is worth checking it against that ruling before we review, rather than after.

Nothing here is a rejection. It is a large and genuinely valuable feature. Thank you for your patience.

Ashishpatra0-0:
@jgupta05072003-code Thank you for the detailed read, and no problem on the wait. Answering your three questions, plus a note on the mastery decision.

1. How `flnLevels.ts` relates to the existing level machinery

Honestly: it is a fourth artifact, not a wrapper or a replacement.

It is a hand-authored map from the legacy 1–59 space to the 9 competency strand names, plus a per-level sub-question count. It does not touch `conceptId`, `levelNumber` or `sCode`, and there is no dependency in either direction between it and `curriculumLevels` / `curriculumMap.ts` / the `skillLevelMap` snapshot. Its own header comment already flags the separation ("Separate id space from `RoleDashboards.tsx`'s 93-entry `FLN_LEVELS_LIST` — don't conflate the two"), and `docs/curriculum-terminology-and-id-mapping.md` does not mention it.

Where it is actually load-bearing: `routes/microPractice.ts` and `services/practiceScheduleService.ts`. Micro-Practice's own progression (`currentLevelId`, `resolved`, via `calculateNextScheduleState` / `getSubsCountForLevel` / `getNextLevelInStrand`) runs entirely in the 59-space and never reads `student.currentLevel`'s 93-scale value. There is also a dead import of it in `index.ts` (`mapCompetencyToLevel` / `KNOWN_COMPETENCIES`, imported but never called) which I can remove.

On resolving it before merge — I looked into this after your earlier review and reported the finding on this PR: the `legacyLevel59` bridge on `curriculumLevels` seeds as `null` for every level, because the reviewed `fln_59_to_93_crosswalk.json` does not exist (only the `.PROPOSED` files). So `getCurriculumLevelByLegacy59()` currently resolves to nothing for all 59 ids, and migrating today would make `getSubsCountForLevel`/`getNextLevelInStrand` silently default rather than fail visibly. Since then, `#423` replaced the crosswalk approach with question-level concept tagging — and in the copy of `questionBank` I can see locally, 0 of 1202 questions currently carry a `conceptId`, so that path is not ready to build against either.

I am happy to do the migration, and agree it should happen. I would just want a steer on which target to migrate onto, given the crosswalk is retired and concept tagging has not been populated yet.

I should also flag against the new doc directly: §6 names `competencyPrerequisites.ts` as the sole owner of advancement logic, and `getNextLevelInStrand` is effectively a second advancement mechanism for a 59-space the doc does not model. That is a real gap, not something I would claim is already compliant.

2. `levels_main.html` (+517/−204)

Agreed this deserves calling out separately, since it affects what gets printed. It is six commits, all Micro-Practice-driven:

- Visible A/B/C labels on match-line questions across ~31 generators, so a teacher grading by hand can see which letter a line points at instead of tracing it.
- Question numbers on vertical arithmetic and fill-in-the-box rows, which previously had none.
- Duplicate-question fixes in several shared generators, including one genuine infinite loop (`compare-equal-match` could be asked for more unique items than its pool could ever produce and would spin forever).
- `generateMicroSet()` and its supporting tables — new, additive, the entry point the backend calls to build a single-section practice set. Does not modify the general worksheet path.
- Visual-confirm reference renderers for 5 drawing-type question types that previously had no gradeable answer.
- A fix for composite generators that silently ignored the requested question count.

It does not touch `buildCleanAnswerKey` or `buildQuestionPaperJSON`, so the general/diagnostic answer-key path is unaffected. The changes to existing shared generators are backward-compatible — output shape is unchanged for the general worksheet path. Happy to walk through any individual generator diff.

3. How practice items are selected

Fully deterministic. No LLM anywhere in the selection or generation path (grepped `gemini`/`openai`/`anthropic`/`generativelanguage` across `microPractice.ts`, `paperGenerator.ts`, `practiceScheduleService.ts`, `flnLevels.ts`, `levels_main.html` — zero hits).

The chain:

1. **Weak competency → level.** Weak competencies come from the latest `EvaluationReport.conceptMastery` filtered to `'Needs Practice'`. `mapCompetencyToLevel()` is a hard-coded lookup to the easiest 59-space level in that strand. No randomness.
2. **Progression.** `calculateNextScheduleState()` is a fixed three-tier rule on the last score: ≥80% advances a sub-level (or to the next level in the strand) and lengthens the interval; 50–79% holds; <50% holds and shortens the interval. Same input, same output, every time.
3. **Content.** The resolved `(levelId, subIdx)` drives `generateMicroSet()`, which looks up that level's hard-coded section list — each section names one of ~130 author-written generator functions and a fixed parameter object (e.g. `sec('A','Add and Match','addition-add-and-match',{matchCount:5,fillCount:6,vertCount:8,maxSum:10})`). Generators randomize only *within* those declared bounds using plain `Math.random()` helpers.
4. **Answer key** is computed by the same call that produced the questions, from the exact values generated — not inferred afterward.

On "what stops an unsuitable question reaching a child": since nothing is generated, there is no filtering step to describe. The property is structural — a human authored every generator and every level's parameter bounds ahead of time, and the only request-time choice is which values inside those bounds get used. The realistic failure mode is a mis-authored level (a bad bound, a params typo), which is a code-review concern for that generator, not runtime moderation.

One boundary worth stating precisely: *identifying* that a child is weak in Fractions may involve Gemini upstream, since `conceptMastery` can be set by `evaluateAIDiagnostic()` during diagnostic grading (with a deterministic fallback when no key is configured). *Selecting which questions get printed* for that weakness never does.

4. On the mastery ruling in #288 / #289

I could not find the decision at those numbers — in this repo #288 is a table-pagination overflow fix and #289 is an admin-creation password-hash fix. I searched commit messages, docs and source for `#288`, `#289` and "server-authoritative" without a match, so I suspect the numbers are off or the record lives somewhere I cannot see. Could you re-share the link?

What I can confirm without it: `PracticeSchedule.resolved` is written only in `calculateNextScheduleState()`, called only from the `/api/practice/submit` handler. There are no client-side writes to it (grepped `MicroPractice.tsx`). Semantically it is narrow — "no further 59-space Micro-Practice content exists in this strand" — rather than a claim about a child's mastery of a concept, and the doc's own Mastery term ("concept status in mastery progression matrix") has no implementation in the codebase yet. So I do not think it currently competes with anything, but I would rather check it against the actual ruling than assert that from inference.

I've also removed the dead `flnLevels.ts` import from `index.ts`. Happy to make any of the other changes above ahead of the review pass

Ashishpatra0-0:
@jgupta05072003-code Separate from the above — something I found while testing, in code outside this PR.

`routes/students.ts` (~line 1141) derives conceptMastery from threshold checks against recommendedLevel :

```
'Number Sense': recommendedLevel >= 15 ? 'Strong' : 'Needs Practice',
'Shapes':       recommendedLevel >= 25 ? 'Strong' : 'Needs Practice',
'Fractions':    recommendedLevel >= 35 ? 'Strong' : 'Needs Practice',
'Operations':   recommendedLevel >= 12 ? 'Strong' : 'Needs Practice'
```
These thresholds look calibrated for the 1–59 scale. `classLevelRange()` puts Class 2 at 43–61, Class 3 at 62–75 and Class 4 at 76–93, and placement uses weakest-level mapping over that range — so `recommendedLevel` cannot fall below 43 for a Class 2 student. Every comparison is therefore always true, and all four strands come back `Strong` for every Class 2, 3 and 4 student regardless of score. I traced a 2/10 diagnostic producing `Strong` across all four.

Downstream effect, and why it's relevant to reviewing this PR: `getWeakCompetenciesForStudent()` filters `conceptMastery` for `Needs Practice`, so most students' weak-competency lists come back empty. Teachers can still generate practice by selecting competencies manually, but the automatic targeting is effectively disabled for Class 2 and above. Worth knowing when testing this branch.

Not proposing a fix — the correct thresholds seem like a curriculum decision rather than a code one.

Minor, unrelated: the `CLAUDE.md` baseline note about two pre-existing type errors (`index.ts` / `paperGenerator.ts`) is stale — both were fixed upstream in July, and `tsc --noEmit` is clean on `main` today. Worth updating that line so it doesn't mislead.

jgupta05072003-code:
Flagging something #433 already said but never reached you: this issue names #306 explicitly as one of two PRs built against the manual-entry / client-side ICR path that is being retired.

> Whether micro-practice papers can go through Gemma like diagnostic papers. If they can, `MicroPracticeAnswerEntry.tsx` becomes a confirm-and-correct view and `pdfjs-dist` is likely unnecessary.

That question is now answered: **yes** — per the D2 ruling on the FLN tracker (12 Sep), every paper in this pipeline goes through Gemma 4 extraction with mandatory human confirm-or-correct, not manual typed entry. #433 also flags that `MicroPracticePaperUpload.tsx:138` calls `POST /api/icr/rasterize-pdf`, which does not exist in `backend/src/` — worth checking before this goes further, since as written that upload path cannot work.

This does not replace the fuller review #306 is queued for; flagging early so the rework, if any, is smaller by the time that review happens.

Ashishpatra0-0:
@jgupta05072003-code I'd replied to both of these on this PR shortly before your comment — they may have crossed.

On `rasterize-pdf` : the route does exist. It was in `index.ts:160` — not moved into `routes/microPractice.ts` during the route-organisation refactor, since it's an ICR utility even though Micro-Practice is its only caller, which makes it easy to miss when looking there. It also doesn't exist on `main`, as it was introduced on this branch, which may be where the "does not exist" read came from. I've now moved it into `routes/microPractice.ts` so it's where you'd expect.

On Gemma: noted, and noted on the reasoning — verification-on-every-answer producing the Gemma accuracy baseline as a byproduct is a stronger argument for it than I'd have made for manual entry. I'll rework `MicroPracticeAnswerEntry.tsx` into a confirm-and-correct view over Gemma's output.

One steer I'd want before starting that. Alongside typed answers, that screen handles a "visual-confirm" answer type I added for drawing-based questions — tally marks, fraction shading, symmetry completion, division pictures. These have no extractable value: the teacher compares the child's drawing against a rendered reference image and confirms yes/no. D2's framing is about every extracted answer being confirmed or corrected, which these never are. Does D2 cover them too, or is a non-extraction confirm path acceptable for that subset? If they're in scope, `pdfjs-dist` likely stays regardless, since those comparisons need the page rendered.

jgupta05072003-code:
This is a large, well-described feature (spaced-repetition micro-practice + evaluation workflow). Before a full review pass, two open architectural questions need addressing since this touches level/difficulty progression and evaluation:

1. **#465 (open)** — "Paper generator still matches legacy `level` instead of the reviewed `mappedLevel`". This PR's difficulty-progression logic generates papers against student level — could you confirm whether it reads the legacy `level` field or the reviewed `mappedLevel`, and if the former, how it avoids the same bug?
2. **#433 (open)** — "Retire the client-side ICR/OCR scanning path; standardise extraction on Ollama Gemma 4". This PR adds new photo/PDF upload + evaluation paths. If these route through the client-side ICR/OCR pipeline that #433 is retiring, we'd rather not add new surface area to something being phased out — could you clarify which extraction path this uses?

No code-quality issues found yet — this is purely about not building on a foundation that's about to move. Once these are clarified, happy to schedule the full review.

Ashishpatra0-0:
@jgupta05072003-code Both fair questions — answering each.

**#465 — neither field.** Micro-Practice doesn't draw from `questionBank` at all. `generateMicroPracticePaper` and `generateMultiCompetencyMicroPaper` launch Puppeteer against `levels_main.html` and call its procedural `generateMicroSet()`; neither contains a single `dbStore` call. The one `questionBank` path in `paperGenerator.ts` is inside `generateDiagnosticPaper` (via `getStudentAssignedQuestions` → `generateClassPaperFromAtlas`), which Micro-Practice never calls. Grepping `routes/microPractice.ts`, `practiceScheduleService.ts`, `paperGenerator.ts` and `levels_main.html` for `questionBank`, `mappedLevel`, `reviewStatus` and `conceptId` returns nothing. So there's no `{ $match: { level } }` join here to get wrong, and no `conceptId` stamping.

I should be straight about *why* it's unaffected, though: not because this PR is on the correct scheme. Progression runs entirely in the legacy 59-space via `flnLevels.ts`, so it sidesteps the join bug by not doing the join. I flagged `flnLevels.ts` as a fourth id space in an earlier comment here, along with why migrating it isn't currently possible — `legacyLevel59` seeds as `null` for every level since the reviewed crosswalk doesn't exist, and per #465's own update that approach is superseded by per-question tagging, which isn't populated yet. The migration target is genuinely unclear rather than merely undone, so I'd want a steer on which to move onto.

**#433 — two halves, and only one is clean.**

The upload path uses QR identification via `jsQR` to determine which paper and student a photo belongs to. No extraction, and #433 explicitly keeps this. It doesn't import `IcrScanner` or touch the client-side ICR machinery. `MicroPracticePaperUpload.tsx:138` calls `/api/icr/rasterize-pdf`, which does exist — it was in `index.ts`, and I've since moved it into `routes/microPractice.ts` so it's findable. Work item 6 is stale.

The answer-entry path is genuinely in scope. `MicroPracticeAnswerEntry.tsx` is manual typed entry, which D2 retires. Accepted — I'll rework it into a confirm-and-correct view over Gemma's output.

One steer before I start. That screen also handles a "visual-confirm" answer type I added for drawing questions — tally marks, fraction shading, symmetry completion, division pictures. These have no extractable value: the teacher compares the child's drawing against a rendered reference image and confirms yes/no. D2's framing is about every *extracted* answer being verified, which these never are. Does D2 cover them, or is a non-extraction confirm path acceptable for that subset? It changes the shape of the rework, and whether `pdfjs-dist` stays.

jgupta05072003-code:
Hi @Ashishpatra0-0, thank you for the detailed answer on the #465 question — that clears up the architecture concern nicely, and it's good to know Micro-Practice runs on its own separate path.

One small thing left before this can move forward: our automated check (Parikshak) shows this PR isn't linked to an issue yet. Could you add "Closes #<issue-number>" to the PR description for whichever issue this feature is meant to resolve? Once that's added, I'll take this up for a full review.

Thank you for your patience on this one — I know it's been about a month.

Ashishpatra0-0:
@jgupta05072003-code Thank you — I've added `Closes #398` to the description.

On why that one: #398 came out of my onboarding document, and this PR had already fixed ~19 generators as part of building the feature, so the remaining work sat naturally here rather than in a separate PR against the same file. I've now completed it — 23 generators fixed (10 that hang, 13 that repeat questions at their production count).

One thing worth flagging: the checker from #513 reports 45 findings, but I verified the other 22 by hand and they're artifacts of the checker's own segmentation rather than broken content. Details and examples are in the updated description. I'd rather say that openly than quietly change 22 generators that produce correct papers.

Ready for the full review whenever you get to it.

PR #163, opened 2026-08-15
feat: Upgrade Regional Performance Analytics Dashboard with Multi-Parameter Filtering

VaibhavChauhan615:
### Overview
This Pull Request upgrades the Geographical and Performance Analytics section inside the teacher and admin panel dashboards. Previously, the Analytics page rendered a static list of schools with dummy progress stats and overlapping layout alignment issues. 

This PR transforms the Analytics panel into an interactive, multi-parameter dashboard that enables administrators to slice data dynamically, visualize student level growth over time, check gender disparities, and trace learner engagement streaks.

### Key Changes
* **Multi-Parameter Filtering**:
  * Added **Time Horizon** selection (`7d`, `30d`, `6m`, `1y`) to scale performance timelines.
  * Added **Grade Band** filter (`Preschool`, `Class 1`, `Class 2`, `Class 3-4`) to slice cohort metrics.
  * Added **Gender** filter (`Male`, `Female`) to isolate demographic analytics.
* **Expanded Metrics Banner (6 Cards)**:
  * Calculates `Total Schools`, `Total Students`, `Avg FLN Level`, and `Certification Rate` (Level 5+ benchmark) dynamically.
  * Added `Active Educators` (teachers count inside the filtered scope) and `Remedial Support Rate` (percentage of students requiring intervention).
* **Advanced Data Visualization**:
  * Integrated a custom, glowing SVG progress line chart showing average student FLN levels over time.
  * Built a strand proficiency breakdown showing mastery levels across different math topics (Number Sense, Operations, Shapes, Measurement, Fractions).
* **Demographic & Leaderboard Insights**:
  * Added Boy-vs-Girl gender disparities split card comparing avg levels and certification rates.
  * Appended a Regional Performance Ranking leaderboard displaying top-performing jurisdictions vs regions needing support.
  * Integrated active learning streaks tracker (tracking 3+ and 5+ day learner streaks).
* **Layout Alignment Fix**:
  * Restructured table grid columns with a fixed-width code column (`w-36`) to resolve the overlapping text overlay bug.

### Verification & Performance
- Verified that all workspace projects typecheck and compile cleanly with `npm run lint` (exit code `0`).
- Calculations use optimized `Set` lookup maps to process metrics in `O(N+M)` time, ensuring the dashboard rendering is completely lag-free.

PR #74, opened 2026-07-19
feat(teacher-workflow): add multi-class student management, diagnostics, performance dashboard and evaluation reports

sreeja37:
# Summary

This PR enhances the **Teacher Workflow** by replacing mock implementations with dynamic backend-driven functionality and improving student management, diagnostics, performance tracking, and evaluation reporting.

### Student Registration
- Integrated newly registered students into the live application workflow.
- Ensured newly created students are immediately reflected across all teacher-facing views.

### Multi-Class & Multi-Section Support
- Added dynamic Class and Section selection across teacher modules.
- Removed assumptions of a single classroom.
- Enabled seamless switching between multiple classes and sections.

### Student List
- Unified the student roster with the Teacher Dashboard.
- Ensured Student List displays the same live data as other teacher views.
- Resolved synchronization issues between the Dashboard and Student List.

### Student Profile
- Added Class and Section selectors.
- Dynamically updates the student list based on the selected class and section.
- Uses a shared student data source to maintain consistency across the application.

### Editable Personal Details
- Added Edit / Save / Cancel functionality.
- Enabled editing of student personal information.
- Added validation and backend persistence.
- Ensured updates are reflected across all teacher-facing views.

### Diagnostic Workflow
- Integrated the diagnostic workflow with backend APIs.
- Improved endpoint routing and workflow handling.
- Replaced mock implementations with live backend integration.
- Fixed issues related to worksheet generation and AI diagnostic flow.

### Performance Dashboard
- Added dynamic filtering by Class, Section, and Student.
- Improved state synchronization.
- Ensured performance data stays consistent with the selected student scope.

### Evaluation Reports
- Replaced mock reports with backend-driven Evaluation Reports.
- Added dynamic filtering by Class, Section, and Student.
- Implemented report aggregation and summary statistics.
- Added defensive handling for empty or partial datasets.
- Fixed runtime rendering and filtering issues.

## Backend
- Added the Evaluation Report module.
- Implemented:
  - Controller
  - Service
  - Repository
  - Interface
  - Model
  - Routes
- Registered Evaluation Report routes.
- Extended student services and APIs to support the updated teacher workflow.

## Frontend
- Added the Evaluation Report service.
- Added a React Query hook for Evaluation Reports.
- Updated teacher panels to consume live backend data.
- Improved state synchronization across Student Profile, Performance, Diagnostics, and Reports.
- Removed remaining mock dependencies where applicable.

## Verification

- ✅ Frontend build successful.
- ✅ Backend build successful.
[...]

PR #49, opened 2026-07-15
feat: add interactive student progress tracking panel component

kundankumar301206:
Implemented a custom student performance summary tracking component inside the frontend application workspace directory. This addition calculates and displays key cumulative learning metrics across literacy and math modules to provide educators with quick student progress visibility at a glance.

PR #307, opened 2026-08-22
feat(teacher-dashboard): Teacher Assessment Dashboard, Visual Progress Tracker & Onboarding Doc

jiyasingh951s-arch:
### Summary of Changes
- **Teacher Diagnostic & Progress Tracker (`TeacherDashboard.tsx`):** Added responsive visual analytics for teachers to monitor student baseline performance, competency distributions, and attendance-score correlations.
- **Student Analytics Hook (`useStudentAnalytics.ts`):** Implemented client-side aggregation logic grouping students into FLN performance tiers (Needs Intervention, Progressing, Proficient).
- **Targeted Remediation Queue:** Added an actionable table to directly identify struggling students below baseline thresholds for targeted worksheet generation.
- **Onboarding Documentation:** Added `Ideas/ONBOARDING-Jiya-Singh.md` covering system audit, architecture gaps, feature proposals, and contribution details.

### 💡 How This Contribution Differs from Existing PRs
- **Dedicated Class-Level Visual Analytics (vs. PR #74 & PR #163):** Focuses specifically on frontline classroom diagnostic charts and attendance-score correlation rather than admin oversight or raw roster CRUD.
- **Aggregate Class Mastery Breakdown (vs. PR #49):** Provides high-level cohort distribution rather than single-student profile cards.
- **Actionable Remediation Pipeline (vs. PR #54 & PR #59):** Directly connects low-score benchmarks (<50%) to instant worksheet assignment triggers for teachers.
- **Self-Contained Client State:** Includes resilient fallback data in `useStudentAnalytics.ts` ensuring clean client-side rendering.

### Visual Preview
image

FLN Assessment Portal - Google Chrome 22-08-2026 03_50_47

### Testing & Verification
- Tested dynamic grade (All, 1, 2, 3) and competency level (10, 16, 25, 31) filtering.
- Verified Recharts responsive bar and scatter chart layouts.
- Confirmed fallback handling for mock datasets.


jgupta05072003-code:
Thank you for this work, and our apologies for the delay in coming back to you. The analytics itself is the useful part here, and the shape of the hook reads well.

Before we can review the logic properly, there are three things that stop the branch building. We checked this by taking the branch and running the type check, so these are not guesses.

**1. `recharts` is not a dependency**

`TeacherDashboard.tsx` imports it, but it does not appear in any `package.json` in the repository and nothing else uses it. This PR does not add it either. It would be helpful if it could be added to `frontend/package.json` with the lockfile change committed alongside.

It may also be worth calling it out in the description. Bringing in a charting library is a decision worth someone agreeing to, rather than something that arrives as a side effect of a feature.

**2. `UNAUTHORIZED_EVENT` does not exist**

`App.tsx` imports it from `./services/apiClient` at line 26 and uses it at lines 79 and 80, but that module only dispatches the string `'fln_unauthorized'` directly. There is no such export, and this PR does not add one. Either exporting the constant from `apiClient.ts`, which would be the tidier option, or going back to the string would both work.

**3. The branch removes dark mode, and we do not think that was intended**

This is the one we would most like you to look at. The diff takes out of `App.tsx`:

- the `isDark` state and its `fln_theme` persistence
- the theme toggle control itself
- the `onNavigateHome` handler

`Layout` still requires all three of `onNavigateHome`, `isDark` and `onThemeToggle`, which is what the type error at `App.tsx:167` is reporting. So this is not only a typing problem: as the branch stands, the theme switch is gone from the interface and the navigate-home action goes with it.

Nothing in the PR description mentions the theme, which is why we read this as a rebase artefact rather than a deliberate choice. If it was deliberate, removing a user-facing feature would be better as its own PR with the reasoning attached, so it can be considered separately from the dashboard work.

**4. There are now two components called `TeacherDashboard`**

Not a blocker, but worth settling before this merges. Yours is at `components/TeacherDashboard.tsx`, takes no props and fetches its own data. The existing one at `components/dashboards/TeacherDashboard.tsx` takes `user`, `token` and `onNavigate`, and is still imported and re-exported at `RoleDashboards.tsx:239-240`. After this PR the older one is still exported but nothing reaches it.

Two live components sharing a name and loading their data differently will be confusing for whoever comes next. Could you let us know whether you are replacing the existing dashboard or adding a second view beside it? If it is a replacement, it would be helpful to remove the old component and its re-export in this same PR. If it is a second view, a distinct name would avoid the ambiguity.

Once it builds we will do a proper pass over the analytics logic, which we have not been able to run yet.

jiyasingh951s-arch:
Good morning @jgupta05072003-code ma'am! 

All four build blockers and naming items have been resolved and pushed in commit `5d451cf4`:

* **Dependencies:** Explicitly added `recharts` to `frontend/package.json` and pushed the updated `package-lock.json`.
* **API Alignment:** Exported `UNAUTHORIZED_EVENT` from `frontend/src/services/apiClient.ts` to satisfy `App.tsx`.
* **Theme & Navigation:** Restored `isDark`, `onThemeToggle`, `onNavigateHome`, and local storage persistence in `App.tsx`.
* **Disambiguation:** Renamed this component to `TeacherAnalyticsDashboard` (`TeacherAnalyticsDashboard.tsx`) to avoid any collision with the base teacher dashboard.
* **Build Status:** `npx tsc --noEmit` compiles cleanly with zero errors.

### UI & Theme Verification

**Dark Mode View:**
WhatsApp Image 2026-09-04 at 5 56 53 AM
WhatsApp Image 2026-09-04 at 5 58 00 AM
WhatsApp Image 2026-09-04 at 5 58 46 AM
**Light Mode View:**
WhatsApp Image 2026-09-04 at 5 57 28 AM

The dashboard is ready for the logic pass. Thank you!

jgupta05072003-code:
Hi @jiyasingh951s-arch — thank you for this, and thank you for following up patiently. The visual direction is genuinely nice: the competency distribution and the attendance-to-performance correlation are exactly the two things a teacher would want to see first, and the Recharts work looks clean.

I would like to raise one thing before we go further, and please take it as a request for a next step rather than a criticism of the work.

**The dashboard is not connected to any real data yet.**

`frontend/src/hooks/useStudentAnalytics.ts` defines a fixed `DEFAULT_STUDENTS` array of ten students (`STU-101 Aarav Sharma`, `STU-102 Ananya Verma`, and so on) with hardcoded scores and attendance percentages, and every chart is computed from that array. `TeacherAnalyticsDashboard.tsx` does not call `apiFetch` or use the auth `token` anywhere.

So at the moment, a teacher who opens this would see the same ten invented children regardless of which school or class they belong to. That is the part we cannot merge — not because the code is wrong, but because the numbers on screen would not be their students' numbers.

The good news is that the wiring should be straightforward, because the endpoints already exist:

- `GET /api/students` — already role-scoped, so a teacher receives only their own roster (see `backend/src/routes/students.ts`)
- `GET /api/classes` — for the class/grade filter

Would you be willing to change `useStudentAnalytics` to accept the real student list (fetched with `apiFetch` and the `token`, the way `TeacherDashboard.tsx` does it) and keep the existing filtering and categorization logic on top of that? The chart code and the NIPUN benchmark buckets can stay exactly as they are — it is really just replacing the data source.

One other thing to be aware of: this PR currently conflicts with `frontend/src/App.tsx` on the latest `main`, so a rebase will be needed as well.

Please do continue with it — the presentation layer is good work and it would be a shame to leave it unfinished. Happy to help if you get stuck on the fetch wiring.

jiyasingh951s-arch:
Hi @jgupta05072003-code Ma’am, I wanted to clarify something regarding PR #307. I’ve been working on the changes based on your feedback, particularly changing the dashboard from individual student-level tracking to class-level learning insights — instead of displaying individual student names, it now highlights which competencies/topics are strong or need attention across the class. I’ve attached two screenshots of the current implementation for reference.
The implementation is almost complete, and I’m currently finishing the remaining parts. While checking the PR, I noticed that the AI review is flagging that it needs to be linked to an existing issue. I went through the relevant issues, but I couldn’t find one that clearly matches the current scope of the PR after these changes.
Could you please recommend which existing issue this work should be aligned with? If there’s a relevant issue, I’ll update the PR accordingly. Thank you!
ChatGPT Classic 09-09-2026 10_57_51
FLN Assessment Portal - Google Chrome 09-09-2026 10_58_01

PR #308, opened 2026-08-23
Add Weak Topic Detector: weak-topics/mistakes/practice-questions endp…

Ayush20051104:
…oints + search fix + seed data


jgupta05072003-code:
Thank you for this. I reviewed the branch at `21c9f70a`. The feature idea is a good one — aggregating topic weakness across a student's reports and surfacing the exact wrong answers is genuinely useful to a teacher.

There is one blocking problem, and it is a serious one.

**The three new endpoints have no authentication.**

`backend/src/index.ts:1625-1645`:

```ts
app.get('/api/evaluation/:studentId/weak-topics', async (req, res) => {
  const weakTopics = await dbStore.getWeakTopicsForStudent(req.params.studentId);
  res.json(weakTopics);
});

app.get('/api/evaluation/:studentId/mistakes', async (req, res) => { ... });

app.get('/api/practice-questions', async (req, res) => { ... });
```

There is no auth check, no role check, and no scope check on any of them. Anyone who can reach the server and supply a student ID gets that child's aggregated weakness profile and the specific questions they answered wrongly. No login required.

This is a children's education platform holding data on roughly 86,000 students. A named child's academic weaknesses being readable by an unauthenticated request is not something we can ship.

The pattern you need is already in the same file, immediately below your code — `/api/analytics` at line 1647 starts with `const user = getAuthUser(req)` and scopes its response by the caller's role. Please apply the same treatment: authenticate the request, then confirm the caller is actually entitled to see *that* student, rather than trusting the `studentId` in the URL.

That last part matters. Adding a login check alone would still let any teacher in any school read any child's record by ID. We already have that exact bug elsewhere in this codebase and are fixing it; please do not add a third instance.

**Other items**

**The PR currently conflicts with `main`** in seven files, including `backend/src/db.ts`, `backend/src/index.ts`, `frontend/src/App.tsx` and `frontend/src/components/Layout.tsx`. It will need a rebase before it can merge.

**The onboarding document does not follow the required convention.** It is committed as `ideas/onboarding.md`. The pinned instructions ask for `Ideas/ONBOARDING-<your-name>.md`. Note also that the lowercase `ideas/` is a separate directory from the existing `Ideas/` on a Linux server, even though they look identical on macOS. Please rename it.

**A question on attribution.** The commits on this branch are authored by "Avani Jain", the PR is raised from the `Ayush20051104` account, and the feature was described to me by Ayush Sharma. I want to make sure the right people are credited — internship completion depends on it. Could you confirm who worked on this? If it was collaborative, please add a `Co-authored-by:` trailer so both names appear in the history.

---

I know a rebase plus an auth fix is real work. The auth issue is not optional, but if the rebase is the blocker, say so here and I will help with it.

PR #430, opened 2026-09-02
fix(analytics): stop school code overlapping school name in row

lakshya-aran:
The Performance Analytics rows render the school code and school name in adjacent flex children. The code span was fixed at w-20 (80px), but real codes like 'AP_GNT_GNT_0P_30' exceed that width and overflow into the name span, producing the visible overlap on the superadmin Analytics tab.

Fix: drop the fixed w-20 / w-24 widths and let the spans size naturally. Code and school-count become whitespace-nowrap + shrink-0 (they never wrap and never squeeze). Name becomes flex-1 + min-w-0 + truncate (takes the leftover space, ellipsis if it would otherwise push the row off-screen). Progress bar gets shrink-0 so it stays 32px.

Verified: tsc --noEmit clean across both workspaces; vite build and esbuild build both succeed (36.8s + 0.9s).

Visual verification: needs a human to confirm in the browser. Separate data-shape issue (every row shows '0%' and '— schools' — the d object lacks certifiedRate/schools fields) is intentionally NOT fixed here; one kind of change per commit.

### DIFF
--- a/frontend/src/components/BulkDiagnosticWorkflow.tsx
+++ b/frontend/src/components/BulkDiagnosticWorkflow.tsx
@@ -329,16 +329,16 @@ export const BulkDiagnosticWorkflow: React.FC<BulkDiagnosticWorkflowProps> = ({ 
       */}
       <div className="bg-white dark:bg-slate-900 border border-zinc-200 dark:border-zinc-700 rounded-2xl p-6 shadow-sm">
         <h3 className="font-display font-medium text-zinc-900 dark:text-white mb-1">
-          Scan &amp; Upload Answer Sheet
+          Upload Scanned Sheet
         </h3>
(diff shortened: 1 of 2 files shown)


lakshya-aran:
[fix(bulk-diagnostic): relabel 'Scan & Upload' to 'Upload Scanned'](https://github.com/vicharanashala/fln/pull/430/commits/f785e1fe9fb3b1a9fc9cb13d86f44b0ad75769c1) 
Renamed the Bulk Diagnostic panel's "Scan & Upload Answer Sheet" heading and button to "Upload Scanned Sheet" so teachers don't think scanning happens in this app.
      2. Swapped the misleading 🖨 printer emoji on the button for an upload icon 📤, and changed the help text from "scan it here" to "upload
    it here" to match.

PR #438, opened 2026-09-06
Offline scan queue

Samruddhi07s:
This offline scan queue feature is used to store the uploaded sheets by any teacher in a low network area.
Following is the step by step process of how it works:
1. Teacher ,particularly in a low network area tries to upload scanned sheets of students .
2.  Initially the low network would stop the upload but with this feature it can store the sheets on local storage and then when the network is connected it uploads it to the database.
3. Thereby ensuring there is no infinite loading or failed uploads.

### Key Changes:
Added IndexedDB-backed offline scan storage.
Added service worker background sync for queued scans.
Added batch sync API with duplicate-safe processing.
Added online/offline state handling to the ICR scanner.
Added sync status badge and user feedback for queued scans.
Preserved the existing online OCR workflow.

### Output:
When the network is restored, we can syn the local storage with the actual database.


jgupta05072003-code:
Hi @Samruddhi07s — thank you for this, and thank you for the demo video. The offline-first idea is genuinely valuable for our context: low connectivity in schools is a real problem, and a background sync queue is the right way to solve it. The service worker and queue design read well.

There is one thing I would like us to fix before merging, and I am sorry we did not flag it earlier.

**`POST /api/v1/scans/sync-batch` currently has no authentication.**

In `backend/src/modules/scans/scan.controller.ts`, the handler does not call `getAuthUser`, and there is no role check or school scoping. Every other route in the repo goes through `getAuthUser` first. As it stands, anyone who can reach the server could POST scan records — including base64 images — directly into the database, without logging in.

Could you please add the same auth pattern the other routes use? Something along these lines:

```ts
import { getAuthUser } from '../../auth';

app.post('/api/v1/scans/sync-batch', async (req, res) => {
  const user = getAuthUser(req);
  if (!user) return res.status(401).json({ error: 'Unauthorized' });
  // ... and please scope the stored scan to the caller (user.schoolId / user.id)
```

Two smaller points, entirely optional:

- There is no size cap on `imageDataUrl`. A per-request limit would protect us from a very large payload filling storage.
- The `if (!Array.isArray(rawScans))` guard can never be false, since the line above already normalizes `rawScans` to an array. Harmless, just dead code.

Everything else in the PR looks good to me. Please take your time with this — once the auth is in, I am happy to merge it.

Samruddhi07s:
> Hi @Samruddhi07s — thank you for this, and thank you for the demo video. The offline-first idea is genuinely valuable for our context: low connectivity in schools is a real problem, and a background sync queue is the right way to solve it. The service worker and queue design read well.
> 
> There is one thing I would like us to fix before merging, and I am sorry we did not flag it earlier.
> 
> **`POST /api/v1/scans/sync-batch` currently has no authentication.**
> 
> In `backend/src/modules/scans/scan.controller.ts`, the handler does not call `getAuthUser`, and there is no role check or school scoping. Every other route in the repo goes through `getAuthUser` first. As it stands, anyone who can reach the server could POST scan records — including base64 images — directly into the database, without logging in.
> 
> Could you please add the same auth pattern the other routes use? Something along these lines:
> 
> ```ts
> import { getAuthUser } from '../../auth';
> 
> app.post('/api/v1/scans/sync-batch', async (req, res) => {
>   const user = getAuthUser(req);
>   if (!user) return res.status(401).json({ error: 'Unauthorized' });
>   // ... and please scope the stored scan to the caller (user.schoolId / user.id)
> ```
> 
> Two smaller points, entirely optional:
> 
> * There is no size cap on `imageDataUrl`. A per-request limit would protect us from a very large payload filling storage.
> * The `if (!Array.isArray(rawScans))` guard can never be false, since the line above already normalizes `rawScans` to an array. Harmless, just dead code.
> 
> Everything else in the PR looks good to me. Please take your time with this — once the auth is in, I am happy to merge it.

Hello @jgupta05072003-code mam!
It's great to know that you liked the feature idea.
About the changes,  I'll do it as soon as possible and also look into the dead code part.
Once again thank you for considering the feature!

Samruddhi07s:
@jgupta05072003-code ma'am,
I have performed the required changes as suggested by you.
- Done the auth in scan.controller.ts so that only authorized users can upload
- Size limit is set to 5MB otherwise it will throw error
- Removed the deadcode as u had mentioned earlier.

jgupta05072003-code:
Thank you for this. The problem you are solving is a real one and it is easy to underestimate — a teacher in a low-connectivity government school losing a batch of scanned answer sheets to a failed upload is exactly the kind of failure that quietly stops people using a tool. Queueing locally and syncing when the network returns is the right shape of solution.

The implementation is thoughtful too. Client-generated `scanUuid` with a server-side duplicate check is the correct way to make retries safe, the 5 MB guard is sensible, and recording `userId` and `schoolId` on each scan is the right instinct.

Two things to adjust before merge. Both are about matching existing conventions rather than anything wrong with your logic.

**1. `backend/src/modules/scans/scan.controller.ts` conflicts with ADR 001.**

`docs/adr/001-backend-structure.md` records a decision that new backend work goes in `backend/src/routes/<domain>.ts` exporting `register<Domain>Routes(app)`, and explicitly lists the layers we agreed not to introduce:

```
modules/
controllers/
repositories/
models/
interfaces/
```

That ADR exists because two PRs independently proposed different layered structures at the same time, and we would have ended up with three conventions in one backend.

**The fix is small**: move the file to `backend/src/routes/scans.ts`. Your exported function is already named `registerScanRoutes(app)`, so it matches the convention exactly as written — only the path and filename need to change.

You may notice `backend/src/modules/vault` already exists and seems to contradict this. It predates the ADR and is a known exception rather than a precedent; please do not use it as a model for new code.

**2. `/api/v1/scans/sync-batch` introduces a version prefix nothing else uses.**

`grep -rn "'/api/v1" backend/src/` returns nothing — every other route in the codebase is mounted at `/api/...`. Versioning the API may well be worth doing, but it should be a deliberate, repo-wide decision rather than arriving on one endpoint. Could you mount this at `/api/scans/sync-batch` for consistency? If you think versioning is worth pursuing, please open an issue and we can discuss it properly.

**One smaller thing, entirely your call.** The duplicate check is global:

```ts
const existing = await dbStore.getScanByUuid(scanUuid);
```

Since the UUID is client-generated, a scan belonging to another school could in principle be reported back as a duplicate. UUID collisions are vanishingly unlikely so this is not urgent, but scoping the lookup to `user.schoolId` would make the endpoint's behaviour easier to reason about.

**A question rather than a request:** `frontend/public/sw.js` registers a service worker, which is new for this app. Service workers persist across deploys and can serve stale assets if their scope or caching is not tightly bounded. Could you say a little in the description about what the worker caches and what it deliberately does not? If it only handles background sync and caches nothing, saying so explicitly would put the question to rest.

Nothing here is a rejection — the feature is genuinely valuable and the hard parts are already right. Ping me once the file has moved and I will re-review promptly.

Samruddhi07s:
Thanks for the detailed review and guidance, @jgupta05072003-code  ma'am!
Got this idea while applying for internships online where the  network issue persisted more than I could tolerate 😅
As per your observations and suggestions, I’ve made the following updates:
-Moved scan.controller.ts to backend/src/routes/scans.ts to adhere to ADR 001.
-Updated the route path to /api/scans/sync-batch across backend and service worker calls.
-Updated the scanUuid check to scope lookups by user.schoolId.
And about the service worker question, I would like to clarify that the service worker is intentionally limited to background upload sync for offline scans. It does not cache the application shell, static assets, or API responses, and it does not serve stale content across deploys. Its only role is to trigger a queued scan upload when connectivity is restored.

Pushed the fixes to the branch and it is ready for review!

jgupta05072003-code:
Verified all three fixes directly: file is now at backend/src/routes/scans.ts with registerScanRoutes(app) matching ADR-001, route mounted at /api/scans/sync-batch (no v1 prefix), duplicate check now scoped via getScanByUuidForSchool(scanUuid, user.schoolId). Confirmed sw.js only listens for the background-sync event and touches IndexedDB — no fetch/cache interception, matches your description exactly. This addresses all the earlier feedback — ready to merge.

jgupta05072003-code:
One more thing — this now has a merge conflict against main (likely from other PRs that just landed). Could you rebase/merge main into your branch? Once that's clean I'll merge right away, the review above stands.

Samruddhi07s:
> One more thing — this now has a merge conflict against main (likely from other PRs that just landed). Could you rebase/merge main into your branch? Once that's clean I'll merge right away, the review above stands.

Thank you for the detailed review,ma'am!
I am currently travelling without laptop...Will rebase as soon as I get a chance!

PR #440, opened 2026-09-06
feat: add pre-OCR scan quality validation gate and UI quality feedback

ryuma69:
## 🎯 Overview

This PR improves the answer-sheet assessment workflow by introducing a deterministic **pre-OCR scan quality validation layer** and fixes a **dark-mode UI inconsistency**.

Before an uploaded answer sheet is sent to the OCR/ICR pipeline, the system checks basic image quality and provides teacher-visible feedback when the scan may produce unreliable extraction.

---

## 💡 Problem

Poor-quality answer-sheet images can negatively affect OCR/ICR accuracy.

Common issues include:

* Very low image resolution
* Excessively dark or bright images
* Low contrast
* Blurry captures
* Incorrect orientation
* Unsupported or invalid image files

Previously, scans could proceed directly to OCR without providing the teacher with an opportunity to correct obvious quality problems.

---

## 🚀 Pre-OCR Scan Quality Validation

A deterministic validation step has been added before OCR processing.

```text
Answer Sheet Upload
        ↓
Scan Quality Validation
        ↓
 ┌──────────┬───────────┐
 │   PASS   │  WARNING  │
 │          │     ↓     │
 │          │  Override │
 └──────────┴─────┬─────┘
                  ↓
                 OCR
```

Each uploaded scan produces a structured quality result containing:

* Overall status: `pass`, `warning`, or `reject`
* Resolution check
* Brightness check
* Contrast check
* Blur check
* Orientation check
* Human-readable reasons
* Whether the result can be overridden

### Validation checks

The validator checks:

* Supported image types: JPEG, PNG, and WebP
* Empty or invalid files
* Maximum file size
* Minimum image dimensions
* Brightness
* Contrast
* Blur
* Orientation using a deterministic heuristic

No Gemini, Ollama, or other AI/LLM call is required for scan-quality validation.

---

## 🖥️ Teacher-Facing Feedback

The existing scan interface now provides immediate feedback before processing.

### 🟢 Pass

The scan appears suitable for OCR and can proceed normally.

### 🟡 Warning

Potential quality problems are shown to the teacher, such as:

> "This scan may affect OCR accuracy because the image is blurry."

The teacher can choose:

* **Retake Scan**
* **Continue Anyway**

Warning-level results require an explicit override rather than silently allowing the scan to continue.

### 🔴 Reject

Scans that fail critical requirements, such as insufficient resolution, are rejected before OCR and the teacher is asked to capture the sheet again.

---

## 🔐 Override Handling

Warning-level scans can be explicitly accepted using:

```text
proceedDespiteQualityWarning: true
```

The quality result and override decision are preserved with the existing submission metadata, providing an audit trail for later analysis.

Rejected scans do not proceed to OCR.

---

## 🌙 Dark Mode Fix

Fixed a UI card that did not correctly respond to the application's dark-mode theme.

The affected component now follows the existing theme behavior consistently with the rest of the interface.

---

## 🧪 Verification

Tests cover the main quality-validation boundary cases, including:

* Valid clear image
* Unsupported image type
* Invalid/empty file
* Too-small image
* Very dark image
* Very bright image
* Low-contrast image
* Blurred image
* Rotated image
* Warning accepted with explicit override
* Rejected scan prevented from reaching OCR

Additional verification was performed through the running application to confirm:

* Clear scans proceed normally
* Warnings are displayed to teachers
* Warning overrides require explicit action
* Rejected scans do not enter OCR
* Existing assessment functionality remains unaffected
* Dark mode renders correctly for the affected card

Also verified with the project's existing lint, test, and build checks.

---

### Added

* Deterministic scan-quality validation
* Structured scan-quality result
* Scan-quality tests
* Teacher-facing quality feedback

### Modified

* Existing answer-sheet upload/OCR flow
* Existing scan UI
* Affected dark-mode card styling

### Explicitly unchanged

* OCR/ICR extraction implementation
* Evaluation and scoring logic
* Misconception detection
* Database schema
* Automatic image enhancement
* Offline scan queue

The implementation is intentionally limited to **detecting poor scan quality, informing the teacher, and preventing clearly invalid scans from reaching OCR**.

---

## 🌟 Engineering Impact

The new validation layer helps prevent poor-quality inputs from silently propagating through the assessment pipeline:

```text
Poor Scan
   ↓
OCR
   ↓
Incorrect Extraction
   ↓
Incorrect Evaluation
   ↓
Incorrect Diagnostic Data
```

The updated workflow is:

```text
Poor Scan
   ↓
Quality Warning / Rejection
   ↓
Teacher Decision
   ↓
OCR
```

This improves the reliability of downstream OCR/ICR processing while preserving teacher control over warning-level scans.
Screenshot 2026-09-06 144752
Screenshot 2026-09-06 191731
Screenshot 2026-09-06 144744
Screenshot 2026-09-06 191748

### DIFF
--- a/Ideas/ONBOARDING-Omar-Shaik.md
+++ b/Ideas/ONBOARDING-Omar-Shaik.md
@@ -1,0 +1,158 @@
+# Contributor Onboarding Document — Omar Shaik
+
+**Contributor:** Omar Shaik  
+**Target Area:** ICR Scanning Pipeline, Quality Assurance & Teacher UX  
+**Contribution (PR):** Pre-OCR Scan Quality Validation Gate and UI Quality Feedback ([Commit `1582bd85`](https://github.com/vicharanashala/fln/commit/1582bd859be463a3b2559baba49738bbd568978f))  
+
+---
+
+## 1. What is FLN?
+
+**FLN (Foundational Literacy and Numeracy)** refers to the core skills every child must master in early primary education (Preschool through Class 4, ages 3–9): reading simple text with comprehension, recognizing numbers, and performing basic arithmetic (counting, addition, subtraction, simple multiplication).
+
+### The Educational Problem
+In government schools across India, assessments like ASER consistently show that many children advance through grades without mastering basic numeracy or reading. The core problems on the ground are:
+1. **Classroom Heterogeneity:** In one Class 3 classroom, one student is still counting on fingers while another is doing two-digit addition.
+2. **Limited Teacher Time:** Teachers often handle multigrade classrooms and heavy admin work. Manually creating 40 different worksheets and hand-grading them every day is not feasible.
+3. **No Student Screens:** Children in these schools do not have tablets or laptops. **All learning and test-taking has to happen on physical paper.**
+
+The FLN platform solves this by diagnosing each student's genuine level on a 93-level math scale, generating printable personalized A4 worksheets, scanning completed papers with a phone camera, and evaluating answers automatically to advance children toward grade-level mastery.
+
+---
+
+## 2. What do you understand by FLN as a system?
+
+The platform connects classrooms to administrators through a role-based workflow tailored to Indian school administration:
+
+- **Students:** The subjects of the platform. They don't log in; their progress is tracked via unique masked IDs (`aadharMasked`), current levels (1–93), and assessment histories.
+- **Teachers & Volunteers:** The primary field operators. Teachers manage class rosters, generate test papers, scan student answer sheets using phone cameras, and review results. Volunteers assist in low-connectivity or under-resourced schools.
+- **School Principals & Admins (Block, District, State):** Monitor regional health, track school progress through the pipeline (Conducted $\rightarrow$ Scanned $\rightarrow$ Evaluated $\rightarrow$ Certified), and identify lagging schools.
+- **Superadmins:** National oversight team (IIT Ropar / Vicharanashala Lab) managing the curriculum framework and review queues.
+
+### End-to-End Workflow
+1. **Generate:** Teacher generates personalized A4 PDF worksheets with student QR codes and answer boxes.
+2. **Administer:** Students write their answers on physical paper.
+3. **Scan & Ingest:** Teacher takes a photo of the completed sheet using the in-app scanner.
+4. **Pre-OCR Quality Check:** The image is inspected before OCR for resolution, lighting, blur, and orientation.
+5. **Evaluate:** Cloud vision OCR (`ollama-gemma4`) extracts student answers, which are scored against answer keys.
+6. **Progress:** The student's level updates, certificates are awarded if the grade benchmark is cleared, or targeted remedial practice is scheduled.
+
+---
+
+## 3. Current State of the Repository — What Has Been Done So Far
+
+Based on my hands-on work with the codebase while building my PR:
+
+- **Frontend:** Single-page app built with **React 19**, **TypeScript**, **Vite**, and **Tailwind CSS**. Features dedicated dashboards for each role under `frontend/src/components/dashboards/` and scanning components like `IcrTwoStageScan.tsx`.
+- **Backend:** **Node.js** and **Express** with routes modularized in `backend/src/routes/` (`evaluation.ts`, `students.ts`, `worksheets.ts`, `auth.ts`, etc.).
+- **Database:** **MongoDB Atlas** with an automatic fallback to a local JSON database (`data/db.json`) for offline development.
+- **Scanning & Evaluation (`ai-services/` & `evaluation.ts`):** Supports camera upload, blue-ink isolation, and cloud vision OCR using Ollama Gemma 4 (`/api/icr/evaluate-cloud`).
+
+---
+
+## 4. Gaps Observed in the Code
+
+While working on the ICR scanning and evaluation flow, I identified the following concrete gaps:
+
+### 1. No Pre-OCR Scan Quality Validation Gate
+- **Where:** `backend/src/routes/evaluation.ts` (inside `/api/icr/evaluate-cloud`) and `frontend/src/components/IcrTwoStageScan.tsx`.
+- **What:** The system took any uploaded image or PDF and sent it directly to the cloud vision OCR service (`ollama-gemma4`) without validating whether the image was actually readable, in focus, or properly oriented.
+- **Why it matters:** In rural government classrooms, teachers use budget smartphones in poor lighting, with shaky hands, angled perspectives, or low resolution. Sending blurry, dark, or inverted images directly to cloud OCR caused hallucinated numbers, misread digits, and corrupted student diagnostic levels, while wasting expensive GPU and API compute on scans destined to fail.
+
+### 2. Lack of Actionable Feedback and Guidance for Teachers
+- **Where:** `frontend/src/components/IcrTwoStageScan.tsx`.
+- **What:** When a scan failed or produced poor OCR results, the UI only showed a generic "OCR failed" error. There was zero indication of *why* it failed.
+- **Why it matters:** Teachers had no way to know whether the room was too dark, the camera was out of focus, or the sheet was held sideways. This created frustration and repeated bad uploads.
+
+### 3. Vulnerability to Corrupt, Undersized, or Unsupported Files
+- **Where:** `backend/src/routes/evaluation.ts`.
+- **What:** The backend lacked pre-flight sanity checks on buffer sizes and file formats before initiating base64 decoding and model payloads.
+- **Why it matters:** Uploading corrupt files or tiny snippets (<5KB) resulted in unhandled backend errors and unnecessary memory allocations.
+
+---
+
+## 5. Ideas for the Project
+
+### 1. Pre-OCR Scan Quality Validation Gate (Implemented in my PR)
+- **What:** A fast, deterministic inspection layer that evaluates raw image buffers for resolution, brightness, contrast, blur, and orientation before triggering OCR.
+- **Why:** Prevents garbage input from reaching the OCR model, saves latency and server costs, prevents inaccurate student placement, and gives teachers immediate feedback on how to fix their photo.
+- **How:** Read image binary headers to check pixel dimensions ($\ge 600\times 600\text{px}$), sample buffer pixels to estimate brightness and contrast standard deviation, measure high-frequency adjacent differences for blur, and check aspect ratio for landscape rotation.
+
+### 2. In-Browser Real-Time Framing & Edge Detection
+- **What:** Add a live camera viewfinder overlay in the browser that highlights paper edges in green only when all 4 corners are visible and the angle is square.
+- **Why:** Helps teachers frame the worksheet correctly before taking the shot, preventing cropped or skewed photos.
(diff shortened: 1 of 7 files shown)


jgupta05072003-code:
Hi @ryuma69 — thank you for this. The pre-OCR scan quality gate is a thoughtful contribution: catching a blurred or badly lit sheet before it enters the OCR pipeline, and giving the teacher a clear Pass / Warning / Reject with an override, is exactly the kind of thing that saves real time in a classroom. `scanQuality.ts` with its accompanying test file reads well, and I appreciate that you kept the OCR and evaluation logic untouched.

Two things I would like to request before merging.

**1. A small but user-visible issue in `CoordinatorRegistration.tsx`**

The page now calls `t('register.title')` and `t('nav.back')`, but neither the `register` nor the `nav` key group exists in `frontend/src/i18n/locales/en/translation.json` or the `hi` file. i18next falls back to rendering the key itself, so the heading on that page would literally read `register.title` and the button would read `nav.back`.

Could you either add those keys to both locale files, or revert those two `t(...)` calls back to the plain strings? Either is fine — the dark-mode classes you added on that page are a genuine improvement and worth keeping.

**2. Could we narrow the diff a little?**

A large part of the PR looks like an editor auto-format with a different Prettier configuration — mainly template literals being re-wrapped. Ignoring whitespace, the change sizes come down quite a lot:

- `MisconceptionFingerprint.tsx`: 542 changed lines → 44
- `Layout.tsx`: 172 → 64
- `backend/src/db.ts`: 203 → 45

The reformatting is harmless in itself, but it makes the real change hard to see during review and it will collide with other open PRs touching the same files. If you could reset the purely-reformatted files back to `main` and keep only your intentional edits, the PR would be much easier for us to merge quickly:

```bash
git checkout origin/main -- frontend/src/components/MisconceptionFingerprint.tsx
```

(and similarly for any other file where the change is only reformatting)

Also, `AGENTS.md` and `.agents/skills/verify/SKILL.md` look like personal tooling configuration — could you drop those from the PR?

Lastly, this currently conflicts with `main` on `backend/src/db.ts`, so a rebase will be needed.

None of this takes away from the feature itself, which I think is a good addition. Thank you for your patience with the review queue.

ryuma69:
Hi @jgupta05072003-code  Ma’am, thank you so much for the detailed review and the kind words about the pre-OCR scan quality gate. I really appreciate the thoughtful response and the detailed report—it made it very clear what needed to be improved.

I’ve addressed the requested changes. The `register.title` and `nav.back` translations have been reverted to plain strings while keeping the dark-mode improvements.

Regarding the large diff, that was caused by my editor extension applying formatting changes, especially around the template literals. I’ve reverted those formatting-only changes and kept only the intentional feature changes.

I’ve also removed the personal tooling files, rebased with the latest `main`, and resolved the `backend/src/db.ts` conflict. The relevant tests and build checks are passing cleanly.

Thanks again for the guidance, Ma’am. I’ll definitely keep a closer eye on what my extensions decide to “improve” next time .

jgupta05072003-code:
Thank you for this — the problem is well chosen and the PR is unusually well put together. Catching an unreadable scan *before* it reaches the OCR pipeline saves the teacher a round trip and saves us an API call, and the tri-state `pass` / `warning` / `reject` design with a teacher override is exactly the right ergonomics: block what is genuinely unusable, warn on the rest, and leave the person in the room in control. You also shipped tests and put the routes in `backend/src/routes/evaluation.ts`, which matches ADR 001. All of that is appreciated.

I do need to raise one substantial finding before this can merge, and I want to show the evidence rather than just assert it.

## Three of the five checks are measuring the compressed file, not the image

`analyzeScanQuality` parses the PNG `IHDR` and JPEG `SOF` headers to get dimensions — that part is correct and genuinely useful. But brightness, contrast and blur are computed by walking the raw buffer:

```ts
// Estimate Luminance, Contrast & Blur from image buffer samples
const step = Math.max(1, Math.floor(imageBuffer.length / targetSamples));
for (let i = 0; i < imageBuffer.length; i += step) { ... }
```

For PNG that buffer is DEFLATE-compressed, and for JPEG it is DCT-quantised and Huffman-coded. In both cases the bytes are close to uniformly distributed and carry no relationship to pixel luminance.

**I tested it rather than guessing.** I generated two 1000×1000 greyscale PNGs with known true mean luminance and ran your analyzer on your branch:

| Test image | True mean luminance | `brightnessScore` | `status` |
|---|---:|---:|---|
| Near-black page | **11.9** | **134** | `pass` |
| Blown-out white page | **247.9** | **134** | `pass` |

Both images scored **exactly 134** and both passed every check. `contrastScore` came back 71 and 70, `blurScore` 83.2 and 74.8 — effectively the same numbers for two images that are visually opposite.

That is the expected result for sampling compressed data: you converge on the statistics of the compressed stream, roughly a mean of ~127 and a standard deviation of ~74, regardless of the picture. Your thresholds (`< 45` underexposed, `> 235` overexposed, contrast `< 15`, blur `< 8`) therefore sit far outside the achievable range and will essentially never fire.

The tests pass because they exercise the arithmetic on synthetic buffers, which is correct as far as it goes — they just cannot catch the fact that the input is not pixel data.

This matters more than a normal bug because the feature would appear to work. Teachers would see a reassuring "Quality: pass" badge on genuinely unreadable scans, which is worse than showing nothing.

## Two ways forward, both fine by us

1. **Decode the pixels.** `sharp` is the usual choice and can produce a downscaled raw greyscale buffer cheaply (`sharp(buf).greyscale().resize(64, 64).raw().toBuffer()`), which is plenty for brightness, contrast and a Laplacian-style blur estimate. It adds a native dependency, so worth a quick check that it is acceptable for our deploy.
2. **Ship the half that works now.** Resolution and orientation are correct, header-only, dependency-free, and already catch real problems — a 300×300 photo or a sideways worksheet. Merging just those two, with brightness/contrast/blur removed and tracked in a follow-up issue, would deliver value immediately with nothing misleading in the UI.

I would happily merge option 2 today if you would prefer to land something now and do the pixel work separately. Your call.

## Two smaller notes

- **`/api/icr/evaluate-cloud` gains a 422.** Existing callers that do not send `proceedDespiteQualityWarning` will start being rejected on warnings. I checked and `IcrTwoStageScan.tsx` is the only caller in the frontend, and you have updated it, so this looks safe — just flagging it as a deliberate behaviour change to an existing endpoint so it is recorded.
- **The `CoordinatorRegistration.tsx` dark-mode fix is unrelated** to scan quality. No need to remove it, but a separate PR for unrelated fixes makes them easier to review and to revert independently.

To be clear, this is not a rejection and it is not sloppy work — the structure, the tests, the route placement and the UX thinking are all good. The compressed-buffer issue is a genuinely easy one to miss. Tell me which of the two options you would like and I will help get it in.

ryuma69:
Thank you for the thorough and constructive review! Testing against synthetic black and white pages was a great catch — it makes complete sense that sampling compressed DEFLATE/DCT streams converges to uniform distribution noise around ~128 and renders the brightness/contrast/blur thresholds meaningless.

We went with **Option 2**:

1. **Header-Only Deterministic Gate**:
   - Removed the compressed-stream brightness, contrast, and blur heuristics.
   - Kept the zero-dependency checks: MIME validation, file size limits (<5KB / >25MB), minimum resolution gate (<600×600px), and landscape-vs-portrait orientation check.
   - Updated both unit tests (`backend/src/scanQuality.test.ts`) and the frontend UI (`IcrTwoStageScan.tsx`) to reflect this clean scope without misleading metrics.

2. **Future Pixel Decoding**:
   - We will track true pixel decoding (evaluating `sharp` vs deploy/CI constraints, or client-side canvas pre-processing) in a dedicated follow-up issue.

3. **PR Hygiene & 422 Note**:
   - Noted and confirmed the 422 behavior on `/api/icr/evaluate-cloud` (handled by `IcrTwoStageScan.tsx` with the override flow).
   - Reverted the unrelated `CoordinatorRegistration.tsx` dark-mode tweaks on this branch to keep this PR 100% focused. The dark-mode fix has been submitted as a separate PR in **#474**.

jgupta05072003-code:
## Decision: please go with option 2 — ship the two checks that work

We have discussed this internally and would like you to **land the resolution and orientation checks now**, and take brightness, contrast and blur out of this PR.

Reasoning, so it does not feel arbitrary:

- **Resolution and orientation are correct, useful and free.** They parse the PNG `IHDR` and JPEG `SOF` headers, need no dependency, decode nothing, and already catch two real failure modes — a 300×300 phone thumbnail and a sideways worksheet. Teachers get value from those today.
- **Adding `sharp` is a bigger decision than one PR.** It is a native binary dependency and it affects our deploy, so we would rather take that call deliberately and separately than have it arrive attached to a feature review.
- **Shipping something that reports "pass" on an unreadable scan is worse than shipping nothing.** Getting the misleading half out of the way now means the working half can go live immediately.

**Concretely, for this PR:**

1. Keep `analyzeScanQuality` with the resolution and orientation checks and their `metrics` (`width`, `height`).
2. Remove the brightness, contrast and blur checks, their thresholds and their metrics fields.
3. Keep the tri-state `pass` / `warning` / `reject` shape and the override flow — that design is good and we want it. Orientation stays an overrideable `warning`, resolution stays a hard `reject`.
4. Keep `POST /api/icr/check-quality` and the guard on `evaluate-cloud`.
5. Trim `scanQuality.test.ts` to the checks that remain.
6. In the UI, show badges only for the checks that actually run. Please do not leave greyed-out Brightness/Contrast/Blur placeholders — an inactive check reads as a passing one.

**Two small things while you are in there**, both optional:

- The `CoordinatorRegistration.tsx` dark-mode fix is unrelated to scan quality. Happy for it to stay if that is easier, but a separate PR would be tidier.
- Your `Ideas/ONBOARDING-Omar-Shaik.md` can stay in this PR — no need to move it.

**On the pixel-based checks**, please open a follow-up issue rather than dropping the idea. Brightness and blur detection is genuinely worth having; it just needs decoded pixels and a dependency decision. Your threshold values and the tri-state design carry straight over, so that work will be much smaller than this PR was. Reference this discussion so the next person does not repeat the compressed-buffer approach.

Thank you for taking the finding well. The structure, the route placement, the tests and the override UX are all good, and that is the part that lasts.

ryuma69:
Thank you, @jgupta05072003-code  Ma’am! I’ve implemented the agreed Option 2 scope in #440 — resolution and orientation are retained, while the compressed-buffer brightness, contrast, and blur checks have been removed.

I’ve kept the tri-state/override flow and updated the tests and UI to reflect only the active checks. The unrelated dark-mode changes have also been separated into their own PR as discussed.

I’ve opened the follow-up issue for the pixel-based checks and referenced this review discussion there. I’ve also added a note directly in `scanQuality.ts` documenting why sampling compressed image buffers is not valid for pixel-level analysis, so future implementations don’t repeat the same approach.

Thanks again for the detailed review and guidance!

jgupta05072003-code:
Verified — pulled the branch and checked scanQuality.ts directly: only resolution and orientation checks remain at runtime, with the deferral note in place. The brightness/contrast/blur mentions in the onboarding doc are historical design narrative, not live code, and the IcrTwoStageScan UI has no leftover greyed-out badges. This addresses the earlier feedback — ready to merge.

PR #462, opened 2026-09-08
fix: display actual school name in principal dashboard

aryanbnrj:
## Summary

The School Principal dashboard was displaying a hardcoded demo school name — **GPS Model Town Ludhiana** — in its header subtitle, regardless of which school the authenticated principal actually belongs to. Any principal logging in from a different school would always see the wrong school identity.

This PR removes that hardcoded string and replaces it with the actual school name fetched from the backend, so every principal (and teacher) sees their own assigned school.

### \rontend/src/components/dashboards/SchoolDashboard.tsx\
- Added \School\ to the type imports.
- Added a \school: School | null\ state variable.
- Extended the existing \etchSchoolData\ function to also call \GET /api/schools\. The backend already scopes this endpoint to \user.schoolId\ for the \school\ role (\ackend/src/routes/schools.ts\), so the first result is always the authenticated principal's school — no backend changes needed.
- Replaced \GPS Model Town Ludhiana (ID: {user.schoolId})\ with \{school.name}\, falling back to \user.schoolId\ while loading or if the fetch fails.

### \rontend/src/components/dashboards/TeacherDashboard.tsx\
- Same fix for the hardcoded \gps-mt-001 Model Town\ string in the teacher dashboard subtitle. The backend scopes \GET /api/schools\ to \user.schoolId\ for the \	eacher\ role as well.
- Falls back to \user.schoolId\ while loading.

No backend API changes were required.

## Testing

| Check | Result |
|---|---|
| \	sc --noEmit\ (frontend TypeScript) | ✅ 0 errors |
| \
pm run lint --workspace @fln/frontend\ | ✅ Passed |
| \git diff --check\ | ✅ Clean (no whitespace errors) |
| Manual diff review | ✅ Only 2 files changed, only targeted lines |

No frontend test suite is configured in this repository.

## Issue

Closes #443

Issue #449, opened 2026-09-07
Production panels use hardcoded school, teacher, and user fallback data


lakshya-aran:
### Description

The shared panel hook contains hardcoded fallback records for:

- 14 schools
- 4 teachers
- 9 users

When an API request returns an empty array or fails, the frontend uses these demo records as if they were current database data.

### Steps to Reproduce

1. Log in with a principal account.
2. Cause the school or teacher request to fail.
3. Open Analytics, Teachers, or Users.
4. Observe that hardcoded records are displayed.
5. Compare the displayed records with MongoDB.
6. Notice that the UI can show schools, teachers, or users that do not belong to the current school.

### Expected Behavior

If API data is unavailable:

- Show a real loading state.
- Show a retryable error state.
- Show an empty state when the API successfully returns no records.
- Never display hardcoded demo data as current production data.

### Actual Behavior

The frontend selects hardcoded fallback records when live API arrays are empty.

### Acceptance Criteria

- [ ] No hardcoded schools render in production.
- [ ] No hardcoded teachers render in production.
- [ ] No hardcoded users render in production.
- [ ] Empty responses do not trigger fallback data.
- [ ] API failures show an error and retry option.
- [ ] Demo records are isolated to tests or development fixtures.
- [ ] The UI clearly distinguishes loading, empty, and error states.

### Likely Files

- `frontend/src/components/panels/usePanelData.ts`
- `frontend/src/components/panels/AnalyticsPanel.tsx`
- `frontend/src/components/panels/TeachersPanel.tsx`

---

jgupta05072003-code:
**Status: still open — confirmed across all three fallback datasets.** `frontend/src/components/panels/usePanelData.ts:32-47` defines `SCHOOLS_FALLBACK` (14 hardcoded schools), a teacher fallback (4 hardcoded teachers), and `USERS_FALLBACK` (9 hardcoded users, including a "Jinal Gupta" superadmin entry). Around line 120-121, the hook falls back to these arrays whenever the corresponding API call returns empty or fails — with no "showing demo data" indicator anywhere in the consuming UI.

**Confirmed real-world impact (per the issue body, and matching #447 for the schools case):** a principal viewing Analytics, Teachers, or Users with a genuinely empty or momentarily-failing API response sees fabricated records that look identical to real production data, with no way to distinguish the two from the UI alone.

**What's left:**
[...]

PR #471, opened 2026-09-11
Fix/school dashboard issues 441 to 450

lakshya-aran:


jgupta05072003-code:
Thank you for this. Ten issues addressed in one branch, each with a written-up `docs/issues/NN-*.md` and a matching `scripts/verify-iN.sh`, is a standard of rigour we do not usually see. The verification scripts in particular are worth calling out — being able to re-run the evidence rather than take the description on trust is exactly what makes a PR this size reviewable at all.

The `db.ts` change (+109/−403) removing demo fallbacks is also very welcome and lines up with the no-fake-data policy in #449.

**First, an apology, because I have made your rebase harder.**

Earlier today I merged two small PRs that turn out to fix two of the same issues you cover here:

| Issue | Your file | Also fixed by |
|---|---|---|
| 3 — school dashboard hardcoded name | `docs/issues/03-school-dashboard-hardcoded-name.md` | **#462**, merged |
| 6 — principal Students sidebar | `docs/issues/06-principal-students-sidebar.md` | **#472**, merged |

I checked for overlapping PRs on the Tenali side before merging and did not do the same here, which is my mistake and not yours. `SchoolDashboard.tsx` and `PanelViews.tsx` will now conflict when you rebase.

**For what it is worth, your versions of both are better**, and I would like to keep them:

- On the school name, #462 takes `schData[0]`. You look up the record matching `user.schoolId` and fall back to `null`, with a comment explaining that this survives a future change to role scoping. That is the more defensive read and the right one.
- On the principal sidebar, #472 adds the `panel === 'students'` branch and nothing else. Yours does the same but as part of a wider `usePanelData` change that adds `schoolsLoaded` / `schoolsError` / `usersLoaded` / `usersError` / `teachersLoaded` / `teachersError`, which is a real improvement on silent failure.

**So when you rebase, please resolve both conflicts in favour of your versions.** You are overwriting a narrower fix with a broader one, which is the correct direction. No need to preserve anything from #462 or #472.

**On the rest of the PR.** At 40 files and +2743/−699 it needs a dedicated review pass rather than being folded into a batch, and it is queued for that rather than blocked. Two things that would help when we get to it, both optional:

1. **A short mapping in the description** from each issue number to the commits or files that address it. Your `docs/issues/` files are excellent, but a reviewer still has to work out which code change belongs to which issue.
2. **If any of the ten are independent of the others**, splitting them out would let us land those immediately while the larger ones get reviewed. Issue 7 (analytics fallbacks) and issue 8 (analytics endpoint scoping) sound like they might stand alone. Entirely your call — I do not want to create busywork, and if they are genuinely interdependent then one PR is right.

Thank you again, and sorry for the duplicated effort on issues 3 and 6. That is on us for not signalling they were being worked on.

jgupta05072003-code:
## Full review — one blocking problem, and it looks accidental

I have now gone through this properly. The intent and the method are excellent, and I want to say that first because the rest of this comment is a problem report.

**What is genuinely good here:** ten real issues, each with a written-up `docs/issues/NN-*.md` and a runnable `scripts/verify-iN.sh`. Issue 8 in particular — the general analytics endpoint not applying school scope for an authenticated principal — is a cross-tenant data exposure, and finding it is a real contribution. Issues 7 and 9, removing demo fallbacks so a principal never sees invented numbers, are exactly the #449 policy.

## Blocking: the branch removes the Aadhaar protections and breaks the build

`backend/src/db.ts` loses 403 lines, and that deletion takes out more than demo fallbacks.

**1. The build no longer compiles.**

```
main            backend  tsc --noEmit  →  0 errors
this branch     backend  tsc --noEmit  →  26 errors
```

Across `tests/aadhaar-hardening.test.ts` (9), `src/routes/students.ts` (7), `src/db.ts` (3), `src/routes/aadhaarDetokenize.ts` (2) and three files under `src/modules/vault/`.

**2. Three Aadhaar fields are gone from the `Student` interface:**

```diff
-  aadharMasked: string;      // Masked identifier only; plaintext Aadhaar is never stored in MongoDB.
-  aadhaarTokenId?: string;   // Opaque token returned by Aadhaar Vault.
-  aadhaarIdentityId?: string; // Deterministic identity id used for duplicate detection.
```

**3. Most seriously, the write-guard is gone.** `AADHAAR_PROTECTED_UPDATE_FIELDS` blocks `aadhaar`, `aadhaarNumber`, `aadhaarTokenId` and `aadhaarIdentityId` from being written through the generic update path, so plaintext Aadhaar cannot reach MongoDB by accident:

```
occurrences at your merge-base (9 Sep):  2
occurrences on this branch:              0
occurrences on main today:               2
```

I checked the merge base deliberately, because a branch that is behind `main` can look like it deletes things it never had. It does not apply here: the guard existed at your branch point and this PR removes it. `dbStore.listLogsByDetailsPrefix`, used by the vault audit history, goes the same way.

**Why this matters more than a normal build break.** These fields carry government identity numbers for named children. The guard is the thing standing between a well-meaning future `updateStudent({ aadhaar: ... })` and plaintext Aadhaar in the database. Its own source comment says it exists to stop exactly that.

**I am confident this was not deliberate.** Nothing in your ten issue documents mentions Aadhaar, none of the verify scripts touch it, and a 403-line deletion is far more than removing demo fallbacks requires. My guess is that `db.ts` was pruned or regenerated somewhat broadly and this came away with it.

## What we need before merge

1. **Restore the three `Student` fields, `AADHAAR_PROTECTED_UPDATE_FIELDS` and its call site, and `listLogsByDetailsPrefix`.**
2. **Get `cd backend && npx tsc --noEmit` back to zero errors.** That is our only build gate on this repo — there is no test job — so a clean typecheck is the bar.
3. **Sanity-check the rest of the 403 lines.** If the Aadhaar machinery went unnoticed, something else may have too. A read back through that deletion with fresh eyes is worth the time.

Please do not feel you need to rebuild the PR. Restoring those pieces should be a contained change.

## Also, and separately: two of your ten are already merged

I merged two small PRs earlier today that fix the same issues, which was my mistake for not checking overlap first:

| Your issue | Also fixed by |
|---|---|
| 3 — school dashboard hardcoded name | **#462**, merged |
| 6 — principal Students sidebar | **#472**, merged |

**Your versions of both are better and I would like to keep them** — you match on `user.schoolId` rather than taking `schData[0]`, and your `usePanelData` change adds real `loaded`/`error` states. When you rebase, resolve both conflicts in favour of yours.

## Suggested sequencing

You are 17 commits behind `main` and will need to rebase anyway. Once you do:

- If **issues 7, 8, 9** can be split into their own PR, I would take that immediately — issue 8 is a live scoping hole and should not wait behind a large review.
- The remainder can then follow with the Aadhaar restoration.

Entirely your call, and I do not want to create busywork. If one PR is easier, one PR is fine.

Thank you for the rigour — the verify scripts made this review far quicker than it would otherwise have been, and issue 8 is a genuinely valuable catch. Ping me when it is ready and I will turn it around quickly.

PR #474, opened 2026-09-12
fix(ui): add dark mode support to coordinator registration

ryuma69:
### Description
Adds missing dark mode Tailwind classes to the Coordinator Registration page (`CoordinatorRegistration.tsx`) for container text, headers, back button, alert banners, and form cards.

### DIFF
--- a/frontend/src/pages/CoordinatorRegistration.tsx
+++ b/frontend/src/pages/CoordinatorRegistration.tsx
@@ -106,30 +106,30 @@ export default function CoordinatorRegistration() {
   };
 
   return (
-    <div className="mx-auto max-w-2xl">
+    <div className="mx-auto max-w-2xl text-slate-900 dark:text-slate-100">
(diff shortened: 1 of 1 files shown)


jgupta05072003-code:
Merged — and thank you for splitting this out of #440 so quickly. That is exactly what I was hoping for, and it means a real fix lands today instead of waiting behind a larger review.

Checked before merging:
- `frontend/src/pages/CoordinatorRegistration.tsx` was one of the few pages with no dark-mode handling; **62 component files already use `dark:`**, so this brings it in line with the established convention rather than introducing a new one.
- Purely additive Tailwind variants — no light-mode class removed, so the existing appearance is unchanged.
- `npx tsc --noEmit` in `frontend/` is clean.

#440 is still open and waiting on the scan-quality change we discussed: keep resolution and orientation, drop brightness/contrast/blur, and open a follow-up issue for the pixel-based version. No rush on it.

PR #476, opened 2026-09-12
fix(ci): handle empty tracking issue list in repo health check

synapticPush:
### 📌 Summary
Fixes 3 latent bugs in `.github/workflows/repo-health-check.yml`:
1. Adds `// empty` fallback to `--jq '.[0].number // empty'` to prevent `"null"` strings when no open issues exist.
2. Adds `[ "$EXISTING" != "null" ]` guard condition.
3. Adds `--force` flag to `gh label create repo-health` so label creation is idempotent and resilient.

### DIFF
--- a/.github/workflows/repo-health-check.yml
+++ b/.github/workflows/repo-health-check.yml
@@ -4,6 +4,9 @@ on:
   schedule:
     - cron: '30 3 * * *'  # daily, ~9:00 AM IST
   workflow_dispatch: {}
+  push:
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


jgupta05072003-code:
Merged — thanks. Noting for the record: this PR doesn't reference a tracked issue (it fixes 3 latent bugs found directly in `repo-health-check.yml` rather than from a filed report — I checked #470/#317/#256, the existing Repo Health Check issues, and none of them cover this). If these bugs were causing visible false positives/negatives in health-check runs, consider filing a short issue documenting them for the historical record; otherwise no action needed.

PR #489, opened 2026-09-13
refactor: remove five dead OCR provider branches from runCloudOcrOnImage

### LINKED ISSUE #485: Remove the five dead OCR provider branches from evaluation.ts (Google/MiniMax/OCR.space/AWS/Azure)
## Summary

Following the D2 ruling (OCR sheet-reading route, tracker + #433): the platform already enforces Ollama Gemma 4 as the sole OCR engine end to end — every request handler in this file rejects any other `provider` value with a 400 before doing anything else, and the frontend never sends anything but `'ollama-gemma4'`. Verified by reading every call site, not just the guards.

What is left is dead code: the shared `runCloudOcrOnImage` helper still contains five full branches for retired providers that can never be reached.

## Where, and the order to delete them in

`backend/src/routes/evaluation.ts`, inside `runCloudOcrOnImage`. Current line numbers on `main`:

| Provider | Lines | State |
|---|---|---|
| Azure Computer Vision | 692–698 | 501 stub |
| AWS Textract | 314–320 | 501 stub |
| OCR.space | 257–313 | Fully implemented, unreachable |
| MiniMax | 184–256 | Fully implemented, unreachable |
| Google Cloud Vision | 116–183 | Fully implemented, unreachable |

**Delete in this order: Azure first, Google last (bottom of the file to the top).** Every one of these numbers is only accurate against the *original* file — deleting a block shifts every line number below it, so if you go top-to-bottom the second block's line numbers will already be wrong. Deleting bottom-to-top means each remaining number is still correct when you get to it. Re-open the file and re-check line numbers before each deletion regardless; do not trust the table blindly after the first cut.

## Confirmed clean — no other cleanup needed

Checked so you don't have to re-derive this:

- **No orphaned helpers.** `getCloudKey` (line 28) resolves keys generically by env-var name (`ICR_CLOUD_API_KEY_<PROVIDER>`) — it is not provider-specific and needs no change.
- **No special imports to remove.** Every dead branch uses only `fetch` and JS built-ins (`JSON`, `Buffer`, `Array`, `String`). No AWS SDK, no axios, nothing at the top of the file exists only for these five.
- **No `.env.example` entries reference these providers.**
- **No frontend `<select>`/`<option>` lists these providers** in `IcrScanner.tsx` or `IcrTwoStageScan.tsx` — both already hardcode `provider: 'ollama-gemma4'`.
- **The stale error message** — *"...or use the local OCR button"*, pointing at a UI element the frontend's own comments confirm was already removed — appears in exactly two places, lines 317 and 695. Both are inside the AWS and Azure stubs respectively, so deleting those branches removes the stale message with them. No separate fix needed.

## What to keep

Leave the final fallback exactly as it is:

```ts
return { status: 400, body: { error: 'Unknown provider: ' + provider } };
```

It is correct and still needed as a safety net if `provider` is ever malformed. Do not touch the `'ollama-gemma4'` branch or anything in the frontend — both are already correct and already the only path.

## Acceptance

- [ ] `grep -n "provider === 'google'\|'minimax'\|'ocrspace'\|'aws'\|'azure'" backend/src/routes/evaluation.ts` returns nothing
- [ ] `npx tsc --noEmit` in `backend/` stays at 0 errors (that is the bar — the repo has no test job)
- [ ] The five ICR routes still work end to end for `provider: 'ollama-gemma4'` — manual smoke test: photograph a real sheet, confirm extraction still runs

aryanbnrj:
## Summary

Removes the five legacy OCR provider branches (Google Cloud Vision, MiniMax, OCR.space, AWS Textract, Azure Computer Vision) from the \
unCloudOcrOnImage\ helper in \ackend/src/routes/evaluation.ts\.

## Root Cause

Following the D2 platform ruling, Ollama Gemma 4 is the sole OCR engine. The route handler rejects every \provider\ value other than \'ollama-gemma4'\ with HTTP 400 *before* entering \
unCloudOcrOnImage\, so all five branches were structurally unreachable. The frontend also hardcodes \provider: 'ollama-gemma4'\ in every call site (\IcrScanner.tsx\, \IcrTwoStageScan.tsx\). The five branches were dead code firing in production only if an attacker bypassed the route guard — which would return 400 anyway.

## Changes

Each provider was removed in a separate commit, bottom-to-top (Azure → AWS → OCR.space → MiniMax → Google) so each deletion's line numbers were still accurate at the time of the edit:

- \
efactor: remove dead Azure Computer Vision branch\ — 501 stub (9 lines)
- \
efactor: remove dead AWS Textract branch\ — 501 stub + stale error message (9 lines)
- \
efactor: remove dead OCR.space branch\ — full implementation (57 lines)
- \
efactor: remove dead MiniMax vision branch\ — full implementation (73 lines)
- \
efactor: remove dead Google Cloud Vision branch\ — full implementation (68 lines)

Total: **216 lines deleted**, 1 file, 0 additions.

The final \'Unknown provider'\ fallback is intentionally preserved as a safety net per the issue spec.

The comment at line 57–58 naming the retired providers is also preserved as documentation.

## Testing

- \grep\ for \provider === 'google'\, \'minimax'\, \'ocrspace'\, \'aws'\, \'azure'\ in \evaluation.ts\ → **0 matches**
- \	sc --noEmit\ in \ackend/\: the 2 errors reported for \evaluation.ts\ (\express\ and \pdf-lib\ missing modules) are **pre-existing on \main\** — confirmed by running the same check against \main\ before branching. No new errors introduced.
- \git diff --check\ → clean
- Diff is exactly **1 file, 216 deletions, 0 additions**

Closes #485


jgupta05072003-code:
Thanks for this — this is the same fix as #505 (removing the five dead OCR provider branches from `evaluate`, closing #485), which was opened around the same time and has just been merged. Closing this one as a duplicate to avoid a conflicting rebase, but the work here was correct and independently arrived at the same fix — appreciated.

PR #497, opened 2026-09-15
feat: extract legacy SVGs out of MongoDB into files, one place going forward

### LINKED ISSUE #498: SVG markup must live in exactly one place: files, never inline in Mongo
`docs/question-authoring-and-assets.md` already documents that inline SVG markup in MongoDB is an anti-pattern the newest authoring-theme system was specifically built to avoid — but two older collections still violate it:

- `questionBank.svgHtml` (1,202 rows) — actively read by `QuestionReviewPanel.tsx` via `dangerouslySetInnerHTML`.
- `levelHtmlTemplates.htmlContent` (38 rows) — read by nothing in the active codebase; pure legacy duplication of files already committed at `FLN SVG HTML files(22-59)/`.

**Ruling**: going forward, SVG/HTML markup for questions lives in exactly one place — files under `frontend/public/assets/svg/`, referenced by id from Mongo. Never inline database content, never a JSON blob duplicate.

**Scope**: extract both legacy sources to files, repoint the one live reader (`QuestionReviewPanel.tsx`) at the file-based version, then clean the now-redundant fields out of the database.

(Filed retroactively — the work was already done in the same session this decision was made; this issue exists so the PR implementing it has something real to link to, per the no-self-invented-PRs rule.)

jgupta05072003-code:
## Summary
Closes #498.

Closes the last gap in "SVG markup lives in exactly one place, never inline in Mongo" — `docs/question-authoring-and-assets.md` already flagged two places this was being violated.

- **`extractQuestionBankSvgs.ts`** — pulls all 1,202 `questionBank.svgHtml` entries out into individual files under `frontend/public/assets/svg/legacy-question-bank/`, keyed by the existing `questionBankId()` identity, plus a `manifest.json`. Reads from the already-committed `data/questionBank.json`, not a live DB connection — the canonical content already sits in git. **Already run — the 1,202 files + manifest are part of this PR.**
- **`QuestionReviewPanel.tsx`** — repointed from `dangerouslySetInnerHTML` on the inline `svgHtml` field to an `` tag against the new per-question file, via `withBase()` since FLN deploys under a `/fln/` subpath in production. This was the *only* place in the active codebase that actually read `svgHtml` back out. Side benefit: removes a `dangerouslySetInnerHTML` call rendering raw DB content into the DOM.
- **`cleanInlineSvgFromDb.ts`** — the actual database cleanup. **Deliberately not run against production by this PR.** Dry-run by default (matching `migrations/questionTemplateIntentV1.ts`'s existing convention), `--apply` to write. Refuses to unset `svgHtml`/`htmlContent` unless it can first verify the file-based replacement actually covers every row being cleaned.

`levelHtmlTemplates.htmlContent` (38 rows) needed no extraction step at all: nothing in the active codebase reads it, and its original source files are already committed at `FLN SVG HTML files(22-59)/`.

**This PR does not touch the production database.** Running `cleanInlineSvgFromDb.ts --apply` against production is a separate, deliberate step for whoever has that access.

## What's actually been verified (not just planned)

Ran the full pipeline end-to-end against an isolated local MongoDB (port 27099, `fln_svg_migration_test` — not production, torn down after) rather than leaving these as untested checkboxes:

- [x] **Found and fixed a real bug this testing caught**: 105 of the 1,202 extracted files were invalid XML (e.g. literal unescaped `>`/`<` in question text like "Insert the correct symbol ( >  <  = )"). The old `dangerouslySetInnerHTML` tolerated this via lenient HTML parsing; the new `` tag requires valid XML and would have silently shown a broken image for all 105. Fixed in the extractor (scoped escaping against the corpus's actual verified tag vocabulary — svg/text/rect/line/circle/path/polygon/polyline/image), verified the fix leaves the other 1,097 byte-for-byte unchanged with zero regressions, re-ran extraction, confirmed all 1,202 now parse as valid XML, confirmed a fixed file serves correctly via a running dev server.
- [x] **Dry run against a real (local, non-production) MongoDB instance**, seeded via the actual `seedQuestionBank.ts`/`seedLevelHtml.ts` scripts: reported the correct counts (1,202 / 38) and, confirmed by direct query, wrote nothing.
- [x] **`--apply` against that same local instance**: confirmed both fields correctly `$unset` on all 1,240 rows combined, document counts and all other fields (answer, level, questionText) unchanged.
- [x] **The safety abort, deliberately triggered**: temporarily hid the manifest to simulate a stale checkout, ran `--apply` again — confirmed it correctly refused to touch anything and left all 1,202 rows intact, which is the actual scenario this check exists to prevent.
- [x] **Found and fixed a second real bug**: the `` `src` was a hardcoded `/assets/...` path, which would have resolved incorrectly once deployed under FLN's `/fln/` production subpath despite working fine in local dev. Fixed via `withBase()`.

## Still needs someone with production access
- [ ] Run `cleanInlineSvgFromDb.ts` (no flag) against production and confirm the reported counts match what's expected there
- [ ] Only once satisfied, run `cleanInlineSvgFromDb.ts --apply` against production

https://claude.ai/code/session_01STUJa4FAha1uLTGPQTeVyN


jgupta05072003-code:
Filed #499 for the root cause of the escaping bug found while testing this PR — the 105 malformed rows in `data/questionBank.json` itself (and the still-live `questionBank.svgHtml` in production). This PR works around it at extraction time; #499 is the actual fix for the source data, separate follow-up work.

